PR #2, opened 2023-05-27
Start a design document

### LINKED ISSUE #1: The Roadmap of the Nushell package manager.
# :warning: Work In Progress :warning: 

projects to check out
- [`kubouch/nuun`: *Nushell Package Manager Exploration*][`kubouch/nuun`]
- [`amtoine/nupm`: *A manager for Nushell packages.*][`amtoine/nupm`]
- [`cargo`: *The Rust package manager*][`cargo`]
- [`rustup`: *The Rust toolchain installer*][`rustup`]
- [`nupac/nupac`: *Nu package manager written entirely in nu & for nu*][`nupac/nupac`]

- [ ] install Nushell without any other requirement -> POSIX compliant / `bash` bootstrap?
- [ ] install `nupm` itself -> see the [`amtoine/nupm` bootstrap]
- [ ] change the version of Nushell with overlays -> [`kubouch/nuun`]

## :wrench: commands to have
- [ ] `nupm new [--script] [--module]`: create a new local package with template files ([`kubouch/nuun`])
- [ ] `nupm init`: initialize an existing local package ([`kubouch/nuun`])
- [ ] `nupm test [--module] [--workspace]`: run unit and integration tests of local package
  \`---> *i think the [`std run-tests`] command just be pulled out of the standard library of Nushell* 
- inspired by [`cargo`]
  - [ ] `nupm install`: ... (different ways of doing this in [`amtoine/nupm`] and [`kubouch/nuun`])
  - [ ] `nupm uninstall`: ...
  - [ ] `nupm update`: ... (prototype in [`amtoine/nupm`])
  - [ ] `nupm bench`: ...
  - [ ] `nupm search`: ...
  - [ ] `nupm publish`: ...
  - [ ] `nupm doc`: ...

## :candy: other nice things to have
- [ ] install plugins
- [ ] have a centralized list of *official* and *community* packages to
  - pull them with `nupm install`
  - search them with `nupm search`
 
## :bulb: concepts to have in mind
- activations from [`amtoine/nupm`]: ...
- overlays from [`kubouch/nuun`]: ...
- require the least amount of user manual intervention
\`---> required things might be
  - using the CLI of `nupm` of course
  - bootstrapping
  - (optional) setting `NUPM_HOME` to change the location of all files (and be XDG-compliant imo)
  - adding a path to `PATH`
  - adding a path to `NU_LIB_DIRS`
  - adding at least one `use` or `overlay use` to bring some commands and modules to the startup scope of the shell
- how do we configure `nupm`
  - config file?
  - environment variables?
  - both?

[`kubouch/nuun`]: https://github.com/kubouch/nuun
[`amtoine/nupm`]: https://github.com/amtoine/nupm
[`cargo`]: https://github.com/rust-lang/cargo
[`rustup`]: https://github.com/rust-lang/rustup
[`nupac/nupac`]: https://github.com/nupac/nupac

[`amtoine/nupm` bootstrap]: https://github.com/amtoine/nupm/blob/main/bootstrap/bootstrap.nu
[`std run-tests`]: https://github.com/nushell/nushell/blob/234ba197d7bd0c7c6aef4b77afff06c5ffd6ae38/crates/nu-std/lib/testing.nu#L308-L437

kubouch:
A WIP of a design document. Still has some parts to fill in

should close #1 by moving all the design into `DESIGN.md` :+1: 

:point_right: see the document [here](https://github.com/nushell/nupm/blob/kubouch-patch-1/docs/design/README.md)


amtoine (commented):
thanks for this @kubouch :innocent: 

maybe there are some missing things compared to #1?

and i'd just vote for
- `package.nuon` instead of `project.nuon`
- `$env.XDG_DATA_HOME` support and default to `~/.local/share/nupm/` instead of cluttering the `$env.HOME` with yet another `~/.nupm/` :wink:

skelly37:
To summarize our meeting with @amtoine @kubouch @Yethal

- We start out with implementing simple `install` & `uninstall` on Linux
- Everyone takes a look at their package managers and see what we would like to actually see here
- We default to the interactive way `[Y/n]` prompt, but allow some kind of `--yes` flag
- We need to add our desired API to the docs (something can be extracted from here): https://github.com/nupac/nupac/blob/main/docs/DESIGN.md
- `install` works by default only with non-installed packages. Reinstall, upgrade or specific revision needs explicit flag for that
- We use the virtual environment idea with the possibility to have global packages to handle both a programming language and a convenient shell
- We stick to some metadata.nuon, keys to consider: https://github.com/nupac/nupac/blob/main/docs/PACKAGE_METADATA.md. Also Nushell version should be addressed
- Dependencies should include versions
- We can use something like: `~/.nupm/packages/<package_name>/<version>/` to handle multiple versions of packages
- We should be able to support multiple package repositories 
- Some kind of way to initialize a new package kinda template, like `nupm new <package_name>`


I'm off today, but I will surely try to add something to the design document on Monday

amtoine:
thanks @skelly37, that looks very complete with what i remember from the meeting :ok_hand:

skelly37:
Also, regarding configuration:

How about using profiles? We won't end up with a ridiculous amount of envs, the users won't be forced to edit json/nuon/whatever by themselves. And one could have multiple configs on their machine, they could be copy-pasted etc...

e.g. `nupm profile set --default interactive_mode=false`

For example conan uses that: https://docs.conan.io/1/reference/profiles.html

amtoine:
i've made some tiny tweaks to the document => the API still is the biggest missing part :ok_hand:

Yethal:
Regarding installation:
I don't think defaulting to git clone as main installation method is a good idea.
* It requires git which may or may not be installed on target system
* It requires packages to be stored in git repositories

IMO installing packages over HTTP is a better choice
* Requires no external dependencies, both download itself and file verification can be done with just nu commands
* Allow packages to be stored in places other than git repositories
* Can still be used to download files from git (the way we currently do it in nupac)

amtoine:
:question: a question regarding this `DESIGN.md` document itself: **do we add some bullet points to keep track of what has been done?**

Yethal:
I mentioned this during the call and then totally forgot about it. I'd like to include a set of helpers for with the module (maybe as a submodule that is not exported by default) to be used in build.nu scripts so that package maintainers do not have to reinvent the wheel and come up with 100 different ways to e.g download a file. Examples of such helpers that come to mind:
* Download a file and verify its checksum
* Assert required binary is available in PATH
* Extract an archive (by detecting and calling an OS-appropriate archive util)
* Attempt to install a dependency using a system package manager
* Clone a git repository in a temporary directory

DESIGN.md
```diff
@@ -0,0 +1,156 @@
...
+## API / CLI Interface
+
+Nushell's module design conflates CLI interface with API -- they are the same.
+
+- `nupm new [--script] [--module]`
+    - create a new local package with template files ([`kubouch/nuun`])
+- `nupm list`
+    - list currently installed packages and if they're out of date
+- `nupm install`
+    - install package into the currently active overlay (can override which overlay to install to)
```
skelly37, line 109: I'd add some info about the 3 possible states and our approach to each of them:

- Installed & up to date
- Installed but outdated
- Not installed

DESIGN.md
```diff
@@ -0,0 +1,156 @@
...
+- `nupm add`
+    - add a dependency to the current project
+    - it is different from `nupm install`: this one adds the dependency to the MANIFEST_FILE, `nupm install` does not
+- `nupm uninstall`
+    - uninstall a package from a currently active overlay (can override which overlay to install to)
+- `nupm update`
```
skelly37, line 115: Are we going to re-fetch the repository cache upon each `update`? 

If yes, we might do it too often. If no, the additional command to refresh it should be added

DESIGN.md
```diff
@@ -0,0 +1,156 @@
...
+## Package repository 
+
+Packages need to be stored somewhere. There should be one central "official" location (see https://github.com/NixOS/nixpkgs for inspiration).
+
+Additionally, user should be able to add 3rd party repositories as well as install local and other packages (e.g., from the web, just pointing at URL),
+as long as it has `METADATA_FILE` telling `nupm` what to do.
+
+## API / CLI Interface
+
+Nushell's module design conflates CLI interface with API -- they are the same.
```
skelly37, line 102: Overall cool, but I'd add expected flags & args already at this point

DESIGN.md
```diff
@@ -0,0 +1,156 @@
...
+    - parse the project to search for errors but do not run it
+- `nupm test`
+    - run unit and integration tests of local package
+- `nupm bench`
+    - run benchmarks
+- `nupm doc`
+    - generate documentation
+- `nupm publish`
+    - publish package to a repository
+    - (not sure about this one, for now, repository can be a github repo with packages submitted by PRs)
```
skelly37, line 131: I think it's too early to think about deployment from CLI. As you've said yourself, the repository will be on github and there's a long journey ahead before we can do anything more sophisticated

DESIGN.md
```diff
@@ -0,0 +1,156 @@
...
+- `nupm install`
+    - install package into the currently active overlay (can override which overlay to install to)
+- `nupm add`
+    - add a dependency to the current project
+    - it is different from `nupm install`: this one adds the dependency to the MANIFEST_FILE, `nupm install` does not
+- `nupm uninstall`
+    - uninstall a package from a currently active overlay (can override which overlay to install to)
+- `nupm update`
+    - update all packages in a currently active overlay overlay (can specify package and/or overlay name)
+    - can be used to self-update (let's avoid having both "update" and "upgrade")
```
skelly37, line 117: Agreed, though I would guard `nupm` auto-update with `--self-upgrade` or sth like that, so it's not a default behavior. One should be able to prevent `nupm` from surprise updates that can break their all workflow

DESIGN.md
```diff
@@ -0,0 +1,156 @@
...
+We might want `nupm`support both types of dependencies.
+
+## Package repository 
+
+Packages need to be stored somewhere. There should be one central "official" location (see https://github.com/NixOS/nixpkgs for inspiration).
+
+Additionally, user should be able to add 3rd party repositories as well as install local and other packages (e.g., from the web, just pointing at URL),
+as long as it has `METADATA_FILE` telling `nupm` what to do.
+
+## API / CLI Interface
```
skelly37, line 100: I would also add some info that `nupm` will be interactive by default (`[Y/n]`) with the standard `--yes (-y)` flag to make it non-interactive

DESIGN.md
```diff
@@ -0,0 +1,156 @@
...
+
+We could later think about being able to extend nupm, like cargo has plugins.
+
+## Other
+
+* activations
+* doc generation
+* test running
+* benchmark running
+* configuration (do not add until we really need something to be configurable, keep it minimal, case study of a project with minimal configuration: https://github.com/psf/black)
```
skelly37, line 154: Maybe something like that:

We will try to avoid configuration until it's really needed. When it will happen, we will most probably choose the idea of having `nupm` profiles.

DESIGN.md
```diff
@@ -0,0 +1,156 @@
+# Design of nupm :warning: Work In Progress :warning: 
+
+This file collects design ideas and directions. The intention is iterate on this document by PRs with discussion.
+
+> **Note**  
+> in the following, until we settle down on precise names, we use the following placeholders:
+> - `METADATA_FILE`: the file containing the metadata of a package,
```
skelly37, line 7: Are we going to prepare some `METADATA_REFERENCE.md` in this PR, too? 

I know, emergent design is cool, but we already know a bit about what can be useful in a package manifest/metadata. I would advise against postponing the metadata design.

DESIGN.md
```diff
@@ -0,0 +1,156 @@
...
+    - list all overlays
+    - `nupm overlay list <overlay-name>` lists all packages installed within the overlay
+- `nupm overlay export`
+    - dump all the installed package names, versions, etc. to a file
+- `nupm overlay import`
+    - create overlay from exported file
+
+We could later think about being able to extend nupm, like cargo has plugins.
+
+## Other
```
skelly37, line 148: We can also explicitly mention (as someone pointed out on Discord) that `nupm` will be cross-platform. I suggest making it exactly like in nupac (with `nupm new` defaulting to the full list, forcing the developer to explicitly remove support for a specific OS):

- android
- linux
- macos
- windows

DESIGN.md
amtoine: i have
- `nupm update nupm`
- `nupm update --self`
- `nupm update --all`

in my `amtoine/nupm` that would all update `nupm` itself :+1:

DESIGN.md
kubouch: Yeah, these are good ones.

DESIGN.md
kubouch: I'd like to avoid the Ubuntu situation where you need to `sudo apt update`, then `sudo apt upgrade`. I think if you're running `update`, you're interested in the latest versions so each package should check for the latest version from the latest version of the repository. Not sure if we need to add separate repository refresh, `cargo update` also works as-is.

DESIGN.md
kubouch: Yes, we can add flags, I just added high-level description of the commands. We can even have a separate file for the commands.

DESIGN.md
kubouch: Agreed, we can explicitly mark it as not intended for now.

DESIGN.md
kubouch: Or `nupm update self`, agreed.

DESIGN.md
kubouch: Yes, we can sketch it out, then implement piece by piece as we need the additional functionality.

DESIGN.md
skelly37: Okay, we can assume that `update` always fetches the newest repo cache. In case of no "freshness", no separate refresh is needed.

DESIGN.md
skelly37: > Or `nupm update self`, agreed.

I would rather see it as a flag than a separate subcommand tbh. @amtoine's suggestion looks perfect t me 😋

DESIGN.md
```diff
@@ -0,0 +1,164 @@
...
+    - install package into the currently active overlay (can override which overlay to install to)
+- `nupm add`
+    - add a dependency to the current project
+    - it is different from `nupm install`: this one adds the dependency to the MANIFEST_FILE, `nupm install` does not
+- `nupm uninstall`
+    - uninstall a package from a currently active overlay (can override which overlay to install to)
+- `nupm update`
+    - update all packages in a currently active overlay overlay (can specify package and/or overlay name)
+    - can be used to self-update (let's avoid having both "update" and "upgrade")
+- `nupm search`
```
skelly37, line 118: For `search` I suggest adding `--unsupported (-u)` or `--all (-a)` flag. By default `search` will show only supported packages (supported nushell version & OS), while `search -u`/`search -a` will show even the ones you are not able to install.

DESIGN.md
skelly37: So to summarize flags I'd suggest:

- `--reinstall (-r)`: reinstall package if installed
- `--update (-u)`: update local packages if outdated

In case of specifying both flags, `--update` should have higher priority imo

DESIGN.md
```diff
@@ -0,0 +1,164 @@
...
+- `nupm new [--script] [--module]`
+    - create a new local package with template files ([`kubouch/nuun`])
+- `nupm list`
+    - list currently installed packages and if they're out of date
+- `nupm install`
+    - install package into the currently active overlay (can override which overlay to install to)
+- `nupm add`
+    - add a dependency to the current project
+    - it is different from `nupm install`: this one adds the dependency to the MANIFEST_FILE, `nupm install` does not
+- `nupm uninstall`
```
skelly37, line 113: I'll just leave one comment but 3 commands need `--yes (-y)` flag to prevent interactive prompt for confirmation (`[Y/n]`)

- `install`
- `uninstall`
- `update`

DESIGN.md
```diff
@@ -0,0 +1,164 @@
...
+We might want `nupm`support both types of dependencies.
+
+## Package repository 
+
+Packages need to be stored somewhere. There should be one central "official" location (see https://github.com/NixOS/nixpkgs for inspiration).
+
+Additionally, user should be able to add 3rd party repositories as well as install local and other packages (e.g., from the web, just pointing at URL),
+as long as it has `METADATA_FILE` telling `nupm` what to do.
+
+## API / CLI Interface
```
skelly37, line 100: It kinda depends on metadata which is to be done buuut...

If we agree on having both short and long package descriptions, all commands displaying some output (`list`, `install`, `search`, `uninstall`, `update`) should by default display the short description and flag (one of them):

- `--long (-l)`
- `--long-desc (-l)`
- `--long-description (-l)`

To display the long description instead of the short one

And I think that this comment concludes the flags discussion for now 😄

DESIGN.md
```diff
@@ -0,0 +1,164 @@
...
+* Add `NUPM_HOME/lib` to NU_LIB_DIRS
+* Add `NUPM_HOME/overlays` to NU_LIB_DIRS
+* Make the `nupm` command available somehow (e.g., `use` inside `config.nu`)
+
+WIP: I have another idea in mind, need to think about it. The disadvantage of this is that the default install location is not an overlay.
+We could make `nupm` itself an overlay that adds itself as a command.
+
+There are several approaches:
+* bootstrap using shell script sourced from web (like rustup)
+* embedded inside Nushell's binary
```
Yethal, line 74: Embedding inside nushell binary would give us installation and upgrading for free. If anything it's the nushell itself that could use a bootstrap script

skelly37:
I thought I will post the metadata.md suggestion here, so it won't got lost on discord 😅 

```markdown
# `nupm`'s package metadata reference
## Required attributes
- `author`: Name of the developer/organization/etc.
- `files`: List of records of files being part of the package. Records reference:
  - `checksum`: SHA256 sum of the file
  - `name`: File path (relative to the package scope)
  - `raw-url`: `GET`table link to the file
  - `supported-os`: Exactly like in the top-level metadata
- `name`: Package name
- `nu-version`: Supported Nushell version. Just like `dependencies` might be either exact version or some matcher like "greater than 1.70"
- `short-description`: Short info about the package, displayed by default
- `supported-os`: Operating systems supported by the package, the most broad possibility: `{"arch": ["*"], "family": ["*"], "name": ["*"]}`. Matched by `$nu.os-info`
- `url`: Package website/GitHub repository. Basically a place where one can find some additional info about the package
- `version`: Version of the package. Semantic versioning is advised to enable users to have more generic requirements
## Optional attribtues
- `dependencies`: Packages needed by the package — versions have to be specified. e.g. `[nupm/0.7.0]`. Semantic versioning is also supported: `[nupm/~0.7]`
- `installer`: Name of a script (relative to the package scope) that will install the package instead (or in addition to) of default `nupm` logic
- `keywords`: List of keywords used by `nupm search` in addition to `name`
- `pre-install-hook`: Name of a script (relative to the package scope) to be executed before installation. Might be simple pre-install message or more complex logic
- `post-install-hook`: Name of a script (relative to the package scope) to be executed after installation. Might be simple post-install message or more complex logic
```

DESIGN.md
amtoine: should be good in 8002bb7a873f156d9f4ac7b67053152f9905708a

DESIGN.md
amtoine: should be good in 5ce5dd3d15f0df6df839d4292b0203d0f89d08ac

DESIGN.md
amtoine: doing this right now :+1:

DESIGN.md
amtoine: done in 6fcebf82c2636bf68e9cecee613a4f8af497f221

DESIGN.md
amtoine: could you give an example of a "profile"? :yum:

DESIGN.md
amtoine: added in `docs/design/references/METADATA.md` in 8fa88e65763350fb622e91db59ff6284b5cdf7dc

DESIGN.md
amtoine: done in 7cecdb5e0ecd39f5459181e2871d65cafe7cf294

DESIGN.md
amtoine: should be good in 345da642362f2e0d259f349c8b1cbe673e12f91f

DESIGN.md
amtoine: what about 345da642362f2e0d259f349c8b1cbe673e12f91f? :wink:

DESIGN.md
skelly37: Sadly, I cannot resolve my own issues... But yeah, it's fine.

DESIGN.md
skelly37: perfect 😋

DESIGN.md
skelly37: You mean what would we like to see specifically in `nupm`'s profile or just a real-life example from some package manager?

DESIGN.md
skelly37: What I was suggesting was: "either `-u` or `-a`" 😅 2 flags doing literally the same (especially in this case) seem to be an overengineering. 

If you cannot decide, I personally prefer the `-u` one 😇

DESIGN.md
skelly37: perfect

DESIGN.md
skelly37: I would replace "give" with "print" but generally speaking, fine 😃

DESIGN.md
skelly37: I think that I agree with @kubouch and @amtoine (Discord) to postpone (or discard completely) the pre-/post- hooks functionality. I'm totally fine with removing it from this file. Besides — 👍🏻

docs/design/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+- `nupm new [--script] [--module]`
+    - create a new local package with template files ([`kubouch/nuun`])
+- `nupm list`
+    - list currently installed packages and if they're out of date
+- `nupm install`
+    - install package into the currently active overlay (can override which overlay to install to)
+    - `--reinstall (-r)`: reinstall package if installed
+    - `--update (-u)`: update local packages if outdated
+    - `-u` > `-r`
```
skelly37, line 117:
```suggestion
    - Both `-u` & `-r` flags might be specified, but updating has higher priority than reinstalling
```

docs/design/README.md
```diff
@@ -0,0 +1,177 @@
...
+as long as it has `METADATA_FILE` telling `nupm` what to do.
+
+## API / CLI Interface
+
+Nushell's module design conflates CLI interface with API -- they are the same. Not all of the below are of the same priority.
+
+> **Note**  
+> commands like `list`, `install`, `search`, `uninstall`, `update`, ..., i.e. should
+> - give short descriptions by default
+> - give long descriptions with `--long-description (-l)`
```
skelly37, line 107:
```suggestion
> - print short descriptions by default
> - print long descriptions with `--long-description (-l)`
```

I already mentioned it in some thread, but since I can do it as a suggestion then let's make your life simpler :D

docs/design/README.md
```diff
@@ -0,0 +1,177 @@
...
+- `nupm new [--script] [--module]`
+    - create a new local package with template files ([`kubouch/nuun`])
+- `nupm list`
+    - list currently installed packages and if they're out of date
+- `nupm install`
+    - install package into the currently active overlay (can override which overlay to install to)
+    - `--reinstall (-r)`: reinstall package if installed
+    - `--update (-u)`: update local packages if outdated
+    - `-u` > `-r`
+    - `--yes (-y)`: do not ask for user confirmation to use `nupm install` in scripts
```
skelly37, line 118:
```suggestion
    - `--yes (-y)`: do not ask for user confirmation
```

I would just remove it, one can have different reasons to disable interactive prompts

docs/design/README.md
```diff
@@ -0,0 +1,177 @@
...
+- `nupm add`
+    - add a dependency to the current project
+    - it is different from `nupm install`: this one adds the dependency to the MANIFEST_FILE, `nupm install` does not
+- `nupm uninstall`
+    - uninstall a package from a currently active overlay (can override which overlay to install to)
+    - `--yes (-y)`: do not ask for user confirmation to use `nupm uninstall` in scripts
+- `nupm update`
+    - update all packages in a currently active overlay (can specify package and/or overlay name)
+    - can be used to self-update: `nupm update nupm`, `nupm update --self` or `nupm update --all` (would update everything)
+    - `--yes (-y)`: do not ask for user confirmation to use `nupm update` in scripts
```
skelly37, line 128:
```suggestion
    - `--yes (-y)`: do not ask for user confirmation
- `nupm update`
    - update all packages in a currently active overlay (can specify package and/or overlay name)
    - can be used to self-update: `nupm update nupm`, `nupm update --self` or `nupm update --all` (would update everything)
    - `--yes (-y)`: do not ask for user confirmation
```

The same case as above

docs/design/README.md
```diff
@@ -0,0 +1,177 @@
...
+- `nupm search`
+    - search package repository (only supported ones by default)
+    - `--unsupported (-u)`: would also list packages that are not supported in the user's system, e.g. due to OS incompatibilities
+    - `--all (-a)`: would list all packages
```
skelly37, line 132:
```suggestion
```

As mentioned in some thread, let's just stick to one flag

docs/design/references/METADATA.md
```diff
@@ -0,0 +1,22 @@
...
+- `supported-os`: Operating systems supported by the package, the most broad possibility: `{"arch": ["*"], "family": ["*"], "name": ["*"]}`. Matched by `$nu.os-info`
+- `url`: Package website/GitHub repository. Basically a place where one can find some additional info about the package
+- `version`: Version of the package. Semantic versioning is advised to enable users to have more generic requirements
+
+## Optional attribtues
+- `dependencies`: Packages needed by the package — versions have to be specified. e.g. `[nupm/0.7.0]`. Semantic versioning is also supported: `[nupm/~0.7]`
+- `installer`: Name of a script (relative to the package scope) that will install the package instead (or in addition to) of default `nupm` logic
+- `keywords`: List of keywords used by `nupm search` in addition to `name`
+- `pre-install-hook`: Name of a script (relative to the package scope) to be executed before installation. Might be simple pre-install message or more complex logic
+- `post-install-hook`: Name of a script (relative to the package scope) to be executed after installation. Might be simple post-install message or more complex logic
```
skelly37, line 22:
```suggestion
```

We can just get rid of it, as mentioned on Discord and in one thread

docs/design/README.md
```diff
@@ -0,0 +1,177 @@
...
+    - `--yes (-y)`: do not ask for user confirmation to use `nupm install` in scripts
+- `nupm add`
+    - add a dependency to the current project
+    - it is different from `nupm install`: this one adds the dependency to the MANIFEST_FILE, `nupm install` does not
+- `nupm uninstall`
+    - uninstall a package from a currently active overlay (can override which overlay to install to)
+    - `--yes (-y)`: do not ask for user confirmation to use `nupm uninstall` in scripts
+- `nupm update`
+    - update all packages in a currently active overlay (can specify package and/or overlay name)
+    - can be used to self-update: `nupm update nupm`, `nupm update --self` or `nupm update --all` (would update everything)
```
skelly37, line 127:
```suggestion
    - can be used to self-update: `nupm update nupm`, `nupm update --self` or `nupm update --all` (the last one would update each `nupm` package, including `nupm`)
```

Let's not leave space for inambiguities 😇

docs/design/references/METADATA.md
amtoine: done in ad6b0cd8a06722ce05f90a2f67bf782f40c87010

docs/design/README.md
amtoine: done in ad6b0cd8a06722ce05f90a2f67bf782f40c87010

docs/design/README.md
amtoine: done in done in ad6b0cd8a06722ce05f90a2f67bf782f40c87010

i've just made the "script" an example

docs/design/README.md
amtoine: done in ad6b0cd8a06722ce05f90a2f67bf782f40c87010

as said below, i've made "to use `nupm install` in scripts" an example, not the only reason

docs/design/README.md
amtoine: done in ad6b0cd8a06722ce05f90a2f67bf782f40c87010

docs/design/README.md
amtoine: done in ad6b0cd8a06722ce05f90a2f67bf782f40c87010

DESIGN.md
amtoine: maybe both?
i'm not familiar with this concept of "profile" :open_mouth:

amtoine: we can also keep that for later :thinking:

skelly37: I know profiles from [C(++) conan](https://docs.conan.io/1/reference/profiles.html)

It's basically a robust wrapping of config files.
```
> conan profile show myprofile1
Configuration for profile myprofile1:

[settings]
os=Windows
arch=x86_64
compiler=Visual Studio
compiler.version=15
build_type=Release
[options]
[tool_requires]
[env]
```

When it comes to the `nupm` I don't think we have something that needs configuration for now. My idea was: "If we will need some configuration, we will most probably use profiles". But if you feel it's too early for that, we can skip that part

amtoine: oooh i see

amtoine: so you'd have commands to
- create / delete profiles
- set configuration inside a profile
- switch between profiles
?

amtoine: i think we can skip that part for now, if that's ok with ya :relieved: 
the configuration part is mentionned at the end of the design, so we shouldn't forget about it :yum:

skelly37: > so you'd have commands to
> 
>     * create / delete profiles
> 
>     * set configuration inside a profile
> 
>     * switch between profiles
>       ?

Exactly. 



> i think we can skip that part for now, if that's ok with ya relieved the configuration part is mentionned at the end of the design, so we shouldn't forget about it yum

Fine, you can close this issue. I spend too much time with conan profesionally to forget about profiles 😅

amtoine: :+1: 

> I spend too much time with conan profesionally to forget about profiles
> 
:laughing:

skelly37 (dismissed):
Can't wait to actually start implementing it, great job everyone with this brainstorming 😋

kubouch:
Some changes I did:
* renamed `NUPM_HOME/lib` to `NUPM_HOME/modules` because modules can be used in many different ways. Also, it's  better to leave `lib` empty because if we add in the future the ability to install 3rd party non-Nushell tools, they could install their libraries there. Better keep Nushell modules separate.
* The overlays need a bit of thinking and experimentation, we can start without it. The main idea is that a project would create an overlay which you can install as a globally-accessible overlay (Python-style global virtual environment)

Two things I don't agree with:
* Commands should have only one description available via `-h` or `help`. Nushell's help system already contains short and long description which should be enough for our purposes.
* The required metadata fields should be absolutely minimal. Most of the currently required fields could be optional and only required for publishing.
I'll address the above concerns in a separate commit

kubouch (commented):
seems good

skelly37:
>     * Commands should have only one description available via `-h` or `help`. Nushell's help system already contains short and long description which should be enough for our purposes.

@kubouch I'm afraid you misunderstood the intent. It was about the description of packages, that will be displayed when doing some actions like search.

docs/design/references/METADATA.md
```diff
@@ -0,0 +1,25 @@
...
+- `dependencies`: Packages needed by the package — versions have to be specified. e.g. `[nupm/0.7.0]`. Semantic versioning is also supported: `[nupm/~0.7]`
+- `installer`: Name of a script (relative to the package scope) that will install the package instead (or in addition to) of default `nupm` logic
+- `keywords`: List of keywords used by `nupm search` in addition to `name`
+
+## Automatically generated, outside of the user-created metadata file
+- `files`: List of records of files being part of the package. Records reference:
+  - `checksum`: SHA256 sum of the file
+  - `name`: File path (relative to the package scope)
+  - `raw-url`: `GET`table link to the file
+  - `supported-os`: Exactly like in the top-level metadata
```
skelly37, line 25: How do you see their auto-generation?

docs/design/references/METADATA.md
```diff
@@ -0,0 +1,25 @@
...
+## Required attributes
+- `name`: Package name
+- `nu-version`: Supported Nushell version. Just like `dependencies` might be either exact version or some matcher like "greater than 1.70"
+- `version`: Version of the package. Semantic versioning is advised to enable users to have more generic requirements
+- `type`: Tells `nupm` how to install the package
+
+## Required attributes for publishing
+- `author`: Name of the developer/organization/etc.
+- `short-description`: Short info about the package, displayed by default
+- `supported-os`: Operating systems supported by the package, the most broad possibility: `{"arch": ["*"], "family": ["*"], "name": ["*"]}`. Matched by `$nu.os-info`
```
skelly37, line 12: I disagree with marking it `for publishing`. We should always validate if the package is actually installable on the current host.

kubouch:
Ah, oops, I misread, I put the long description back

docs/design/references/METADATA.md
kubouch: Does every file really needs those? I think raw url to the project root (with package.nuon) and supported os for the whole project should be enough.

docs/design/references/METADATA.md
kubouch: I'm thinking about a scenario where you start with `nupm new` and install it only locally: You write it on your machine so it implicitly supports your OS, you don't need to specify that. Looking at `cargo init` as a reference, it also has only three fields. Also, Nushell is cross-platform, we should encourage that by assuming project is cross-platform. If your project is not cross-platform, then you can fill in the `supported-os`.

docs/design/references/METADATA.md
Yethal: This assumes package manifests and the actual package files are stored in the same place which may not be the case for e.g binary packages where only the manifest iself may bo stored in nupm repo but the binary files might be downloaded from somewhere else like github releases

docs/design/README.md
```diff
@@ -0,0 +1,200 @@
...
+* Add `NUPM_HOME/modules` to NU_LIB_DIRS
+* Add `NUPM_HOME/overlays` to NU_LIB_DIRS
+* Make the `nupm` command available somehow (e.g., `use` inside `config.nu`)
+
+> :warning: **WIP**  
+> The disadvantage of this is that the default install location is not an overlay. We could make `nupm` itself an overlay that adds itself as a command, so that you can activate/deactivate it. We might need a few attempts to get to the right solution.
+
+There are several approaches:
+* bootstrap using shell script sourced from web (like `rustup`)
+* embedded inside Nushell's binary
```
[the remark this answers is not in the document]
amtoine: as proposed by @kubouch, let's start simple here and then, as the project becomes more mature, we'll see to move it to main `nushell/nushell` :relieved:

docs/design/references/METADATA.md
amtoine: if there is still discussion about this, i propose to move to either
- an [issue](https://github.com/nushell/nupm/issues/new)
- a [discussion](https://github.com/nushell/nupm/discussions/new?category=ideas)
- the [`#package-manager` channel](https://discord.com/channels/601130461678272522/1112829613920505956) on Discord

PR #6, opened 2023-06-11
FEATURE: add a command to install a local Nushell package

amtoine:
# Description
this PR implements a start for the `nupm install` command:
- installs only local packages, i.e. the files have to be pulled with things like `git` or downloading archives before hand
- defaults to installing the package in the root of the repo
- can be changed with the `--path` option in the case the package is in a subdirectory
- assumes the *package* file exists and is valid
- installs the package by simply copying the files from the repo to the local home of `nupm`

# Note for reviewers
in order to try this with eveything enabled, i use the following commands
```nu
use /path/to/nupm/nupm/
$env.NU_LOG_LEVEL = $env.LOG_LEVEL.DEBUG
$env.NUPM_HOME = ($nu.temp-path | path join "nupm")
```
and then try
```nu
nupm install
# or
nupm install --path .
```
from [`amtoine/nu-git-manager`](https://github.com/amtoine/nu-git-manager)
and
```nu
nupm install --path nu-zellij/
# or
nupm install --path nu-zellij
```
from [`amtoine/zellij-layouts`](https://github.com/amtoine/zellij-layouts) which contains the `nu-zellij` package

### DIFF
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -2,7 +2,7 @@ use std log
 
 # install a Nushell package
 export def main [
-    --path: path  # the path to the local source of the package (default to the current directory)
+    --path: path  # the path to the local source of the package (defaults to the current directory)
 ] {
     let path = ($path | default . | path expand)
 
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -8,6 +8,62 @@ def nupm-home [] {
     )
 }
 
+def throw-error [
+    error: string
+    text?: string
+    --span: record<start: int, end: int>
+] {
+    let error = $"(ansi red_bold)($error)(ansi reset)"
+
+    if $span == null {
+        error make --unspanned { msg: $error }
+    }
+
+    error make {
+        msg: $error
+        label: {
+            text: ($text | default "this caused an internal error")
+            start: $span.start
+            end: $span.end
+        }
+    }
+}
+
+def open-package-file [path: path] {
+    let package_file = ($path | path join "package.nuon")
+
+    if not ($package_file | path exists) {
+        throw-error $"package_file_not_found(ansi reset):\nno 'package.nuon' found in ($path)"
+    }
+
+    let package = (open $package_file)
+
+    log debug "checking package file for missing required keys"
+    let missing_keys = (
+        [
+            [key required];
+
+            [$. true]
+            [$.name true]
+            [$.version true]
+            [$.description true]
+            [$.license true]
+        ] | each {|key|
+            if ($package | get --ignore-errors $key.key) == null {
+                $key
+            }
+        }
+        | where required
+        | get key
+    )
+
+    if not ($missing_keys | is-empty) {
+        throw-error $"invalid_package_file(ansi reset):\n($package_file) is missing the following required keys: ($missing_keys | str join ', ')"
+    }
+
+    $package
+}
+
 def prepare-directory [directory: path] {
     rm --recursive --force $directory
     mkdir $directory
@@ -31,9 +87,8 @@ export def main [
     --path: path  # the path to the local source of the package (defaults to the current directory)
 ] {
     let path = ($path | default . | path expand)
+    let package = (open-package-file $path)
 
-    # NOTE: here, we suppose that the package file exists and is valid
-    let package = ($path | path join "package.nuon" | open)
     log info $"installing package ($package.name)"
 
     let destination = (nupm-home | path join $package.name)
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -8,6 +8,19 @@ def nupm-home [] {
     )
 }
 
+def prepare-directory [directory: path] {
+    rm --recursive --force $directory
+    mkdir $directory
+}
+
+def copy-directory-to [destination: path] {
+    let source = $in
+    ls --all $source | where name != ".git" | each {|it|
+        log debug ($it.name | str replace $source "" | str trim --left --char "/")
+        cp --recursive $it.name $destination
+    }
+}
+
 # install a Nushell package
 export def main [
     --path: path  # the path to the local source of the package (defaults to the current directory)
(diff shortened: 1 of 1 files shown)


nupm/install.nu
```diff
@@ -0,0 +1,56 @@
...
+        log debug $"destination: ($destination)"
+        cp --recursive $path $destination
+
+        if ($destination | path join ".git" | path exists) {
+            log debug $"removing .git directory from ($destination)"
+            rm --recursive ($destination | path join ".git")
+        }
+    } else {
+        # TODO: add support for updating / reinstalling a package in that case
+        throw-error $"package ($package.name) is already installed"
```
Yethal, line 54: Instead of throwing an error maybe we could just log that package is already installed to make the operation idempotent

amtoine: should be good in 3ff7e91 :relieved: 

i'm curious to know what *idempotent* means here :yum:

Yethal: Run the same command multiple times, get same result every time

amtoine: oooh i see :ok_hand: 

i'm not sure how it applies here because
- on the first `nupm install`, the package would be installed
- on the next ones, `nupm install` would give a warning

is this *idempotent*?
and if it's not, how can we make it so? :open_mouth:

Yethal: I don't think we even need a warning here. If we define the installation command as 'I want these packages installed on my system' and not 'I want to perform installation operation on these packages' (so expression of desired state and not operations to perform) then it's only logical that installing packages that are already present on the system is noop

amtoine: oooooh i see what you mean now, thanks :+1:

nupm/install.nu
```diff
@@ -0,0 +1,43 @@
...
+
+def copy-directory-to [destination: path] {
+    let source = $in
+
+    log info "copying directory"
+    log debug $"source: ($source)"
+    log debug $"destination: ($destination)"
+
+    ls --all $source | where name != ".git" | each {|it|
+        log debug ($it.name | str replace $source "" | str trim --left --char "/")
```
Yethal, line 24: char path_sep instead of "/"

nupm/install.nu
```diff
@@ -0,0 +1,43 @@
...
+}
+
+# install a Nushell package
+export def main [
+    --path: path  # the path to the local source of the package (defaults to the current directory)
+] {
+    let path = ($path | default . | path expand)
+
+    # NOTE: here, we suppose that the package file exists and is valid
+    let package = ($path | path join "package.nuon" | open)
```
Yethal, line 36: Error handling (non-existent package file, corrupted package file, lack of file permissions etc) here

nupm/install.nu
amtoine: good catch :+1: 
i thought i was in a URL probably :eyes:

amtoine: should be fixed in 38e3ba6

nupm/install.nu
amtoine: in this PR, with the goal of being simple, i was inspired by [this message](https://discord.com/channels/601130461678272522/1112829613920505956/1117440843289198643) of @kubouch 
> I think we're conflating two things here a bit:
> 1. Fetch the package from the internet
> 2. Install the package
>
> The step 2 would be identical for any project regardless of where it comes from. The difference between a git package and a local package would be in the step 1.

this means that this PR implements step 2!
so i assume here that the *package* is there and valid, hence the note and the missing checks.

**do you think we should add them here?**
yeah probably, because if you clone the project and then `nupm install` locally, if the *package* file check is only done in the firt stage, you might try to install a broken package :thinking:

amtoine: @Yethal 
i've added a check for the existence and the validity of the `package.nuon` file in e015a28 :+1: 

## examples
- when `package.nuon` does not exist
```bash
Error:   × package_file_not_found:
  │ no 'package.nuon' found in /home/amtoine/.local/share/git/store/github.com/amtoine/nu-git-manager
```
- when the file is missing required keys
```bash
Error:   × invalid_package_file:
  │ /home/amtoine/.local/share/git/store/github.com/amtoine/nu-git-manager/package.nuon is missing the following required
  │ keys: description, license
```

nupm/install.nu
```diff
@@ -0,0 +1,98 @@
...
+    let missing_keys = (
+        [
+            [key required];
+
+            [$. true]
+            [$.name true]
+            [$.version true]
+            [$.description true]
+            [$.license true]
+        ] | each {|key|
```
Yethal, line 51: I think this could be simplified to `keys|all {|key| $key in ($package|columns)}`

amtoine: i've greatly simplified the *package* file key check in 8ab9e69 :yum: 

however, i went with `where` instead of `all` and `columns` because `columns` only returns the list of depth-1 cell-paths :thinking: 
and if we want to test for more complex cell paths... it's more versatile to use `where` and `get -i` :yum:

nupm/install.nu
```diff
@@ -0,0 +1,84 @@
...
+        log debug ($it.name | str replace $source "" | str trim --left --char (char path_sep))
+        cp --recursive $it.name $destination
+    }
+}
+
+# install a Nushell package
+export def main [
+    --path: path  # the path to the local source of the package (defaults to the current directory)
+] {
+    let path = ($path | default . | path expand)
```
Yethal, line 75: shouldn't this be $env.PWD?

nupm/install.nu
```diff
@@ -0,0 +1,84 @@
+use std log
+
+def nupm-home [] {
+    $env.NUPM_HOME? | default (
+        $env.XDG_DATA_HOME?
+        | default ($nu.home-path | path join ".local" "share")
```
Yethal, line 6: Can't we store this in $nu.default-config-dir?

nupm/install.nu
amtoine: ooh yeah i thought about it a while back :+1:

amtoine: should be good in 9351571

nupm/install.nu
amtoine: do you mean having `nupm-home` always return `$nu.default-config-dir`? :thinking: 

i'm not sure these are *config* files, so i chose the XDG location, but i can change the default location, however, i think we should let the user change the location with something like `NUPM_HOME` :yum:

amtoine: should be good in ccf0329 now :wink:

Yethal (approved):

amtoine:
let's land this and have our first `nupm` command :rocket:

PR #7, opened 2023-06-25
add an option to `nupm install` to install scripts

amtoine:
this PR adds a `--script` option to `nupm install`.

- `nupm install --script` requires the use of the `--path` option because installing the `$env.PWD` as a script is ambiguous
- for now `nupm install --script` directly copies the script inside `nupm-home`
- and makes the script executable

# example
from [`silicakes/zellij-sessionizer`](https://github.com/silicakes/zellij-sessionizer), one can install the main Nushell script with something like
```nushell
nupm install --script --path zellij-sessionizer.nu
```

### DIFF
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -67,7 +67,26 @@ def copy-directory-to [destination: path] {
 # install a Nushell package
 export def main [
     --path: path  # the path to the local source of the package (defaults to the current directory)
+    --script: bool  # install the package at --path as an executable script
 ] {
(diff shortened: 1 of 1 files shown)


amtoine:
as agree in the [`#nupm` channel](https://discord.com/channels/601130461678272522/1112829613920505956/1122559005185802360) on the Discord server, the installation should not rely on the user knowing if the package is a module or a script, but rather read the `package.nuon` file :+1:

PR #8, opened 2023-06-29
do not install the Git directory of a package

amtoine:
i've noticed, in case the local package has been pulled via Git, that `nupm install` would install the `.git/` directory.

this is because `name != ".git"` is a weak test given the fact that the paths are absolute because `source` is an absolute path in `copy-directory-to`.

this PR removes the `.git/` directory from the copied items by checking it more strongly.


skelly37 (approved):
lgtm

amtoine:
thanks for the review @skelly37 :relieved:

PR #9, opened 2023-06-29
throw an error when `--path` is not raised

amtoine:
in the future, we will be able to `nupm install <name of package>`, so i think it does not make much sense to default `nupm install` to `nupm install --path .

and as we do not support installing a package by name or URL, i propose to throw an error when running `nupm install` with the `--path` option :+1: 

## inspiration
```
> cargo install
error: Using `cargo install` to install the binaries for the package in current working directory is no longer supported, use `cargo install --path .` instead. Use `cargo build` if you want to simply build the package.
```

### DIFF
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -68,7 +68,10 @@ def copy-directory-to [destination: path] {
 export def main [
     --path: path  # the path to the local source of the package (defaults to the current directory)
 ] {
-    let path = ($path | default $env.PWD)
+    if $path == null {
+        throw-error "`nupm install` requires a `--path`"
+    }
+
     let package = (open-package-file $path)
 
     log info $"installing package ($package.name)"


skelly37 (commented):
I don't quite get what's going on with this `--path`. Wouldn't we simply like to do `install` a package into a default dir (with optional override)? Or maybe I said the truth but you agreed that we do it incrementally and start out with local installation and the package registry will be added in the future?

amtoine:
> I don't quite get what's going on with this `--path`. Wouldn't we simply like to do `install` a package into a default dir (with optional override)? 
>
this is what `nupm install` does right now :yum: 
it installs everything in `nupm-home`, e.g. `~/.local/share/nupm/` on my system.

> Or maybe I said the truth but you agreed that we do it incrementally and start out with local installation and the package registry will be added in the future?
>
yeah i think we're on the same page, i've been doing things incrementally :relieved: 
`nupm install` can only install packages whose source is available locally through the `--path`  option :+1: 
it's like `cargo install --path .` when you're in Nushell's source to install it :wink:

skelly37 (approved):
Okay, seems clear, thanks @amtoine

amtoine:
great, thanks again for the review @skelly37 :relieved:

PR #11, opened 2023-07-06
remove the parentheses in `let` declarations

amtoine:
related to nushell/nushell#9589.

because the new syntax is imo better, i propose to remove parentheses in simple `let` variable declarations, e.g.
```nushell
let package_file = ($path | path join "package.nuon")
```
becomes
```nushell
let package_file = $path | path join "package.nuon"
```

### DIFF
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -26,19 +26,17 @@ def throw-error [
 }
 
 def open-package-file [path: path] {
-    let package_file = ($path | path join "package.nuon")
+    let package_file = $path | path join "package.nuon"
 
     if not ($package_file | path exists) {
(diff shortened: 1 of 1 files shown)


skelly37 (commented):
I personally like the parentheses and their ability to visually separate (to the extent where I do `1 + (2 * 3)` 😅 )

I will not argue about `let package = open $package_file`, but I think that visual separation of pipelines (e.g. `let missing_keys = $required_keys | where {|key| ($package | get -i $key) == null}`) is beneficial.

No explicit approve/disapprove, just a different point of view to compare pros and cons 😇

amtoine:
well, to move forward with this, as we have a :+1: from @fdncred, a neutral from you @skelly37 and an approval from me @amtoine, i'll solve the conflicts and land this :relieved:

PR #15, opened 2023-09-05
Add main nupm command; Start dirs module

kubouch:
This PR adds the top-level `nupm` command, which prompts creating `$env.NUPM_HOME` if it doesn't exist. `$env.NUPM_HOME` is considered to be the default install location of nupm packages and should be always set—either by the user manually, or automatically by nupm itself.

I'm mostly feeling around the code, trying to add small changes first. I'm planning to do a few smaller refactors, eventually bringing some ideas from [`nuun`](https://github.com/kubouch/nuun) I had in mind into nupm.

### DIFF
--- a/nupm/mod.nu
+++ b/nupm/mod.nu
@@ -1,0 +1,13 @@
+use utils/dirs.nu [ DEFAULT_NUPM_HOME nupm-home-prompt ]
+
+export-env {
+    if 'NUPM_HOME' not-in $env {
+        $env.'NUPM_HOME' = $DEFAULT_NUPM_HOME
+    }
+}
+
+export def main [] {
+    nupm-home-prompt
+
+    print 'enjoy nupm!'
+}
--- a/nupm/mod.nu
+++ b/nupm/mod.nu
@@ -3,7 +3,7 @@ use utils/dirs.nu [ DEFAULT_NUPM_HOME nupm-home-prompt ]
 export-env {
     # Ensure that $env.NUPM_HOME is always set when running nupm. Any missing
     # $env.NUPM_HOME during nupm execution is a bug.
-    $env.NUPM_HOME = ($env.NUPM_HOME? | $DEFAULT_NUPM_HOME)
+    $env.NUPM_HOME = ($env.NUPM_HOME? | default $DEFAULT_NUPM_HOME)
 }
 
 # Nushell Package Manager
--- a/nupm/mod.nu
+++ b/nupm/mod.nu
@@ -3,9 +3,7 @@ use utils/dirs.nu [ DEFAULT_NUPM_HOME nupm-home-prompt ]
 export-env {
     # Ensure that $env.NUPM_HOME is always set when running nupm. Any missing
     # $env.NUPM_HOME during nupm execution is a bug.
-    if 'NUPM_HOME' not-in $env {
-        $env.'NUPM_HOME' = $DEFAULT_NUPM_HOME
-    }
+    $env.NUPM_HOME = $env.NUPM_HOME? | $DEFAULT_NUPM_HOME
 }
 
 # Nushell Package Manager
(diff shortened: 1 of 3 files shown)


nupm/utils/dirs.nu
```diff
@@ -0,0 +1,39 @@
+# Directories and related utilities used in nupm
+
+# Decault installation path for nupm packages
+export const DEFAULT_NUPM_HOME = ($nu.default-config-dir | path join "nupm")
+
+# Prompt to create $env.NUPM_HOME if it does not exist and some sanity checks.
+export def nupm-home-prompt [] {
+    if 'NUPM_HOME' not-in $env {
+        error make {
```
amtoine, line 9: i would add a `--unspanned` here, otherwise, the error is less easy to read :confused: 
e.g. for the next `error make` as well
```
Error:   × Root directory ~/.config/nushell/config.nu exists, but
  │ is not adirectory. Make sure $env.NUPM_HOME points at a
  │ validdirectory and try again.
    ╭─[/home/amtoine/documents/repos/github.com/amtoine/nushell-nupm/nupm/utils/dirs.nu:15:1]
 15 │         if ($env.NUPM_HOME | path type) != 'dir' {
 16 │             error make {
    ·             ─────┬────
    ·                  ╰── originates from here
 17 │                 msg: ($"Root directory ($env.NUPM_HOME) exists, but is not a"
    ╰────
```

nupm/utils/dirs.nu
```diff
@@ -0,0 +1,39 @@
...
+            msg: "Internal error: NUPM_HOME environment variable is not set"
+        }
+    }
+
+    if ($env.NUPM_HOME | path exists) {
+        if ($env.NUPM_HOME | path type) != 'dir' {
+            error make {
+                msg: ($"Root directory ($env.NUPM_HOME) exists, but is not a"
```
amtoine, line 17: as you can see in the thread above, the words are missing some spaces with this string concatenation :wink:

nupm/mod.nu
```diff
@@ -0,0 +1,16 @@
...
+    # Ensure that $env.NUPM_HOME is always set when running nupm. Any missing
+    # $env.NUPM_HOME during nupm execution is a bug.
+    if 'NUPM_HOME' not-in $env {
+        $env.'NUPM_HOME' = $DEFAULT_NUPM_HOME
+    }
+}
+
+# Nushell Package Manager
+export def main [] {
+    nupm-home-prompt
```
amtoine, line 13: a question about that though :thinking: 

- the `NUPM_HOME` variable is always set thanks to the `export-env` block above :+1: 
- there is a check when running `nupm` on its own
- but what about running `nupm install`, will the Nupm directory have the same kind of checks? :thinking:

nupm/mod.nu
```diff
@@ -0,0 +1,16 @@
+use utils/dirs.nu [ DEFAULT_NUPM_HOME nupm-home-prompt ]
+
+export-env {
+    # Ensure that $env.NUPM_HOME is always set when running nupm. Any missing
+    # $env.NUPM_HOME during nupm execution is a bug.
+    if 'NUPM_HOME' not-in $env {
+        $env.'NUPM_HOME' = $DEFAULT_NUPM_HOME
+    }
```
amtoine, line 8: quick note: i think this could be simplified to
```nushell
$env.NUPM_HOME = $env.NUPM_HOME? | $DEFAULT_NUPM_HOME
```
:yum: 

not a requested change here though :relieved:

amtoine (changes requested):
i have two things with the errors :yum: 

the rest is basically the same as before but done slightly differently, that's ok :+1: 

thanks for working on Nupm :pray:

nupm/mod.nu
kubouch: Eventually, every `nupm` subcommand (that needs it) will have the NUPM_HOME check, I just wanted to keep this PR small, I'll do another one for `nupm install`.

amtoine: ooh ok gotcha :+1:

amtoine (approved):
lgtm, thanks @kubouch 

looking forward to the follow up PRs :innocent:

PR #18, opened 2023-09-12
Refactor `nupm install`

kubouch:
This PR mostly focuses on `nupm install` which got thoroughly refactored. See also notes in commit messages.

Approximate list of changes:
* Adds `nupm` command itself
* `$env.NUPM_HOME` is always set
* Adds a prompt to create `$env.NUPM_HOME` if it doesn't exist
* Moves directory handling to its own utils module
* Don't require description and license for `nupm install` (only necessary for publishing)
* Fixes module package to install only the module, not the entire project directory
* Module packages can have the "scripts" fields, just like script packages
* Format to fit 80 lines (Sorry! I don't handle too long lines well, as I tend to have multiple vertical windows open at the same time, I think the code looks more neat when it's on shorter lines.)
* Names the installation folders `scripts` and `modules`, depending on which type of project you're installing.
* Custom packages (with build.nu) are `cd`ed to a temporary directory as a "sandbox".
* Installing script package without "scripts" field will try to install `<package_name>.nu` by default.
* Other misc refactors?

### DIFF
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -145,11 +145,12 @@ def install-path [
             }
 
             let tmp_dir = tmp-dir build --ensure
-            cd $tmp_dir
 
-            nu $build_file ($pkg_dir | path join 'package.nuon')
+            do {
+                cd $tmp_dir
+                nu $build_file ($pkg_dir | path join 'package.nuon')
+            }
 
-            cd -
             rm -rf $tmp_dir
         },
         _ => {
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -1,9 +1,5 @@
 use std log
 
-def nupm-home [] {
-    $env.NUPM_HOME? | default ($nu.default-config-dir | path join "nupm")
-}
-
 def throw-error [
     error: string
     text?: string
@@ -65,7 +61,7 @@ def copy-directory-to [destination: path] {
 }
 
 def install-scripts [path: path, package: record<scripts: list<path>>]: nothing -> nothing {
-    let nupm_bin = nupm-home | path join "bin"
+    let nupm_bin = $env.NUPM_HOME | path join "bin"
     mkdir $nupm_bin
 
     for script in $package.scripts {
@@ -97,7 +93,7 @@ export def main [
 
     match $package.type {
         "module" => {
-            let destination = nupm-home | path join $package.name
+            let destination = $env.NUPM_HOME | path join $package.name
 
             prepare-directory $destination
             $path | copy-directory-to $destination
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -33,7 +33,7 @@ def open-package-file [path: path] {
     let package = open $package_file
 
     log debug "checking package file for missing required keys"
-    let required_keys = [$. $.name $.version $.description $.license $.type]
+    let required_keys = [$. $.name $.version $.type]
     let missing_keys = $required_keys | where {|key| ($package | get -i $key) == null}
     if not ($missing_keys | is-empty) {
         throw-error $"invalid_package_file(ansi reset):\n($package_file) is missing the following required keys: ($missing_keys | str join ', ')"
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -72,7 +72,7 @@ def install-scripts [
 
         if ($tgt_path | path type) == file and (not $force) {
             throw-error ($"Script ($src_path) is already installed as"
-                + $" ($tgt_path)")
+                + $" ($tgt_path). Use `--force` to override the package.")
         }
 
         log debug $"installing script `($src_path)` to `($scripts_dir)`"
@@ -110,7 +110,8 @@ def install-path [
             }
 
             if ($destination | path type) == dir {
-                throw-error $"Package ($package.name) is already installed"
+                throw-error ($"Package ($package.name) is already installed."
+                    + "Use `--force` to override the package")
             }
 
             cp --recursive $mod_dir $module_dir
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -162,7 +162,7 @@ def install-path [
 
 # Install a nupm package
 export def main [
-    name    # Name, path, or link to the package
+    package # Name, path, or link to the package
     --path  # Install package from a directory with package.nuon given by 'name'
     --force(-f)  # Overwrite already installed package
 ]: nothing -> nothing {
@@ -172,5 +172,5 @@ export def main [
         throw-error "`nupm install` currently requires a `--path` flag"
     }
 
-    install-path $name --force $force
+    install-path $package --force $force
 }
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -1,5 +1,7 @@
 use std log
 
+use utils/dirs.nu nupm-home-prompt
+
 def throw-error [
     error: string
     text?: string
@@ -88,6 +90,7 @@ export def main [
     }
 
     let package = open-package-file $path
+    print $package
 
     log info $"installing package ($package.name)"
 
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -83,15 +83,16 @@ def install-scripts [path: path, package: record<scripts: list<path>>]: nothing 
 
 # Install a nupm package
 export def main [
-    --path: path  # the path to the local source of the package (defaults to the current directory)
+    name    # Name, path, or link to the package
+    --path  # Install package from a path given by 'name'
 ] {
     nupm-home-prompt
 
-    if $path == null {
-        throw-error "`nupm install` requires a `--path`"
+    if not $path {
+        throw-error "`nupm install` currently requires a `--path` flag"
     }
 
-    let package = open-package-file $path
+    let package = open-package-file $name
 
     log info $"installing package ($package.name)"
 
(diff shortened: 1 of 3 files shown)


kubouch:
Still plan to add tests but otherwise it's done.

nupm/install.nu
```diff
@@ -23,107 +23,161 @@ def throw-error [
...
+        | where {|key| ($package | get -i $key) == null}
     if not ($missing_keys | is-empty) {
-        throw-error $"invalid_package_file(ansi reset):\n($package_file) is missing the following required keys: ($missing_keys | str join ', ')"
+        throw-error (
+            [
+                $"invalid_package_file(ansi reset):"
+                ($"($package_file) is missing the following required keys:"
+                    + $" ($missing_keys | str join ', ')")
+            ]
+            | str join)
```
amtoine, line 51: i think this should `str join` on `char nl` to preserve the newline in the original error, right?

nupm/install.nu
```diff
@@ -23,107 +23,161 @@ def throw-error [
...
-    log debug $"source: ($source)"
-    log debug $"destination: ($destination)"
+# Install list of scripts into a directory
+#
+# Input: Scripts taken from 'package.nuon'
+def install-scripts [
+    pkg_dir: path        # Package directory
+    scripts_dir: path    # Target directory where to install
+    --force(-f)          # Overwrite already installed scripts
+] {
```
amtoine, line 64: i would add a type annotation here
```suggestion
]: list<path> -> nothing {
```

nupm/utils/dirs.nu
```diff
@@ -37,3 +40,35 @@ export def nupm-home-prompt [] {
...
+    let d = $env.NUPM_HOME | path join modules
+
+    if $ensure {
+        mkdir $d
+    }
+
+    $d
+}
+
+export def tmp-dir [subdir: string, --ensure] {
```
amtoine, line 64: for consistency with the two other ones :wink: 
```suggestion
export def tmp-dir [subdir: string, --ensure]: nothing -> path {
```

nupm/install.nu
```diff
@@ -23,107 +23,161 @@ def throw-error [
...
     }
 
     let package = open $package_file
 
     log debug "checking package file for missing required keys"
-    let required_keys = [$. $.name $.version $.description $.license $.type]
-    let missing_keys = $required_keys | where {|key| ($package | get -i $key) == null}
+    let required_keys = [$. $.name $.version $.type]
+    let missing_keys = $required_keys
+        | where {|key| ($package | get -i $key) == null}
```
amtoine, line 43: just curious about that one, you do not find
```nushell
let missing_keys = $required_keys | where {|key| ($package | get -i $key) == null}
```
easy enough to read?

amtoine: ooh, it's about the 80-char width you talk in the PR description :+1:

nupm/install.nu
```diff
@@ -23,107 +23,161 @@ def throw-error [
...
+
+    if not $path {
+        throw-error "`nupm install` currently requires a `--path` flag"
+    }
+
+    if $force {
+        install-path --force $name
+    } else {
+        install-path $name
+    }
```
amtoine, line 182: i think this can be simplified to
```nushell
install-path --force $force $name
```
if you change `install-path` to take `--force (-c): bool = false` :yum:

nupm/install.nu
```diff
@@ -23,107 +23,161 @@ def throw-error [
...
+                    --span (metadata $pkg_dir | get span))
             }
 
-            nu ($path | path join 'build.nu')
+            let tmp_dir = tmp-dir build --ensure
+            cd $tmp_dir
+
+            nu $build_file ($pkg_dir | path join 'package.nuon')
+
+            cd -
```
amtoine, line 152: i think not using `cd -` is safer in such cases, what about
```nushell
do {
    cd $tmp_dir
    nu $build_file ($pkg_dir | path join 'package.nuon')
}
```
instead? :innocent:

nupm/install.nu
```diff
@@ -23,107 +23,161 @@ def throw-error [
...
+                rm --recursive --force $destination
+            }
+
+            if ($destination | path type) == dir {
+                throw-error $"Package ($package.name) is already installed"
+            }
 
-            prepare-directory $destination
-            $path | copy-directory-to $destination
+            cp --recursive $mod_dir $module_dir
```
amtoine, line 114: is this not copying the whole module to the destination as before? :thinking:

amtoine (changes requested):
quite a big refactor, so quite a big review :eyes: 

i'll just have to try `nupm install` with my usual packages to make sure everything is working fine :+1: 

## point by point
> * Adds `nupm` command itself
>
:ok_hand: 

> * `$env.NUPM_HOME` is always set
>
:ok_hand: 

> * Adds a prompt to create `$env.NUPM_HOME` if it doesn't exist
>
~~:question: i did not see that in the changes, can you point me to it?~~
see https://github.com/nushell/nupm/issues/18#issuecomment-1724004360

> * Moves directory handling to its own utils module
>
:ok_hand: 

> * Don't require description and license for `nupm install` (only necessary for publishing)
>
:ok_hand: 

> * Fixes module package to install only the module, not the entire project directory
>
:question: see review threads

> * Module packages can have the "scripts" fields, just like script packages
>
:ok_hand:

> * Format to fit 80 lines (Sorry! I don't handle too long lines well, as I tend to have multiple vertical windows open at the same time, I think the code looks more neat when it's on shorter lines.)
>
:ok_hand: 

> * Names the installation folders `scripts` and `modules`, depending on which type of project you're installing.
>
:question: i will have to try that

> * Custom packages (with build.nu) are `cd`ed to a temporary directory as a "sandbox".
>
:ok_hand: 

> * Installing script package without "scripts" field will try to install `<package_name>.nu` by default.
>
:ok_hand: i see it's in the `default` in the `script` arm of the `match`

> * Other misc refactors?
>
:ok_hand:

nupm/install.nu
```diff
@@ -23,107 +23,161 @@ def throw-error [
...
+export def main [
+    name    # Name, path, or link to the package
+    --path  # Install package from a directory with package.nuon given by 'name'
+    --force(-f)  # Overwrite already installed package
+] {
+    nupm-home-prompt
+
+    if not $path {
+        throw-error "`nupm install` currently requires a `--path` flag"
+    }
```
amtoine, line 176: you do not use `--path` now, so i do not think this guard nor this option is useful now :wink:

kubouch:
@amtoine 
> ❓ i did not see that in the changes, can you point me to it?
Check calls to the `nupm-home-prompt` command at the beginning of every `main`

nupm/install.nu
kubouch: Currently, it doesn't support installing from pipeline input, but it would be a cool addition. I'll change it to `nothing -> nothing` for now.

nupm/install.nu
kubouch: Yeah, it's me being hardcore 80 chars.

nupm/install.nu
kubouch: Unfortunately, it doesn't. It would be super useful for flag passing, though.

nupm/install.nu
kubouch: It's copying the subdirectory containing the module, not the top directory.

nupm/install.nu
kubouch: It's an error to indicate that `--path` is required my `nupm` to function currently. It doesn't support `nupm install some-package` at the moment.

nupm/install.nu
kubouch: Good catch!

nupm/install.nu
kubouch: The `$path` is unused after the check because `--path` is currently the only install option. But as we add more, we'd add a check for `$path` and other types of installation.

kubouch:
During my testing, `nupm install` worked fine, so I'll open this for merging. I'm preparing another PR for adding tests, but I thought it would be better as a separate PR to keep this one shorter because I'm planning to introduce `nupm test` (analogous to `cargo test`).

nupm/install.nu
amtoine: done in https://github.com/nushell/nupm/pull/18/commits/983a645c1d999399f64f00c385963e163cd19193

nupm/install.nu
amtoine: mm i'm not sure i understand here :thinking: 

the command basically does
```nushell
each {|script|
    let src_path = $pkg_dir | path join $script
    let tgt_path = $scripts_dir | path join $script

    ...
}
```
where each `script` is a `string`, so this means that `install-scripts` takes a `list<string>` as input, right? :open_mouth: 

and when we look at [one of the calls to `install-scripts`](https://github.com/nushell/nupm/pull/18/commits/49c4cf5fc11aeeb028b14c5a2701e90213b8fbfa#diff-030e1728e6809d86850afe7153e0c5d902e00f53645fb6c44dda91da9256b49dR119) as `list<string>`, a.k.a. `$package.scripts` is given to its input :yum:

nupm/utils/dirs.nu
amtoine: done in https://github.com/nushell/nupm/pull/18/commits/49c4cf5fc11aeeb028b14c5a2701e90213b8fbfa

nupm/install.nu
amtoine: mm not sure this is correct anymore with https://github.com/nushell/nushell/pull/10456, right? :smirk: :yum:

nupm/install.nu
amtoine: done in https://github.com/nushell/nupm/pull/18/commits/0ae5405d10c9244f63a26b743df9373a096e1b9d

nupm/install.nu
amtoine: ooh yeah, smart, it used to `ls` everything and copy the whole set of files, here `$mod_dir` is the subdirectory which should contain all the module files :+1:

amtoine:
i still have concerns about the design choices here, i propose we switch to the [*module representation design* thread](https://discord.com/channels/601130461678272522/1153371989449375936) on our Discord :yum:

nupm/install.nu
kubouch: Done

amtoine: in https://github.com/nushell/nupm/pull/18/commits/f4a5e427394006da9302e6c64e88aa760a705974 :wink:

nupm/install.nu
amtoine: oooh ok, i think i finally got it :scream:

i was confused by the change but yeah, we installing locally by using the `name` as either the name of the remote package or a path on disk :+1: 

i think this is what confused me... :confused: 
```nushell
    name    # Name, path, or link to the package
```
maybe that would be less confusing to use
```nushell
    package    # Name, path, or link to the package
```
and then `--path` would switch from installation from the remote packages to a local one?

amtoine: @kubouch (just making sure GitHub will ping you :wink:)

amtoine (commented):
i just have a last suggestion but apart from that i'm ready with these changes, if you can just let me a moment to try and fix all my packages and see how it feels :innocent:

nupm/install.nu
kubouch: OK, I'll do that, `package` is clearer, I think. 

Yes, `--path` would force `nupm` to look for the package in your file system instead of remote.

amtoine: yeah, when i read `name`, i thought it was only the "name" of the package in some package index somewhere :open_mouth:

kubouch: Done in https://github.com/nushell/nupm/pull/18/commits/377d6ea93fc073d5a607a4d8ea4719018f33b54f

kubouch: Done in [`377d6ea` (#18)](https://github.com/nushell/nupm/pull/18/commits/377d6ea93fc073d5a607a4d8ea4719018f33b54f)

amtoine (approved):
my bad, i wasn't on the computer last night when i reviewed so i couldn't approve :scream: 

thanks @kubouch for working on this, and i'm glad we managed to converge to something great after some back and forth :pray: 
now i like the new structure and with patches to a bunch of different types of packages, let's see how it goes :wink:

PR #19, opened 2023-09-22
Add `nupm test`

kubouch:
Adds a simple test runner to `nupm`.

`nupm test` requires a `tests` directory module—all exported commands from the module are treated as tests to run. It is possible to filter tests by name. The `--show-stdout` flag is useful for debugging.

Example tests are added for `nupm install`. As a consequence, a few bugs in `nupm install` were discovered and fixed as well.

The three tests included in this PR run ~5.7x faster with `nupm test` than using `std testing run-tests` (~127 ms vs. ~735 ms).

PR #20, opened 2023-10-03
throw better errors

amtoine:
related to
- #19 

~~wait for #19 to land because it introduces more calls to `throw-error` and moves the definition.~~

## description
the `throw-error` command supports giving both an `error` and a `text`, however this would work only with a `--span`...

this PR makes it possible to pass `error` and `text` without a `--span` and updates all the call to `throw-error` to split the `error` and the `text` :+1: 

```
Error:   × missing required option
  │ `nupm install` currently requires a `--path` flag
```

```
Error:   × package_file_not_found
  │ no 'package.nuon' found in /home/amtoine
```

```
Error:   × invalid_package_file
  │ /home/amtoine/documents/repos/github.com/amtoine/
  │ nushell-nupm/package.nuon is missing the following
  │ required keys: name, version, type
```

### DIFF
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -10,7 +10,11 @@ def throw-error [
     let error = $"(ansi red_bold)($error)(ansi reset)"
 
     if $span == null {
-        error make --unspanned { msg: $error }
+        if $text == null {
+            error make --unspanned { msg: $error }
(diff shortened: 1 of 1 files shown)


kubouch:
Thanks!

Issue #22, opened 2023-10-09
`install-scripts` error


KAAtheWiseGit:
Another error during `use nupm` is:

```Nushell
Error: nu::parser::extra_positional

  × Extra positional argument.
     ╭─[/home/kaathewise/fork/nupm/nupm/install.nu:105:1]
 105 │                 $package.scripts
 106 │                 | install-scripts $pkg_dir (script-dir --ensure) --force $force
     ·                                                                          ───┬──
     ·                                                                             ╰── extra positional argument
 107 │             }
     ╰────
  help: Usage: install-scripts {flags} <pkg_dir> <scripts_dir>
```

`install-scripts` uses the `--force(-f): bool`, which I think stopped taking arguments after one of the recent updates.

KAAtheWiseGit:
Same for `install-path`.

KAAtheWiseGit:
Ah, got it!  I use a cut release, not the one installed via Cargo, so I am behind the HEAD.

amtoine:
> Ah, got it! I use a cut release, not the one installed via Cargo, so I am behind the HEAD.
>
yup makes sense :yum: 

we do not have a proper version system for Nupm right now, so you might want to lag behing the latest `HEAD` of `nushell/nupm` if you use the *stables* releases of Nushell :wink:

PR #24, opened 2023-10-09
add dev templates

amtoine:
related to 
- https://github.com/nushell/nupm/issues/22

## description
because Nupm and Nushell are quite tied together, i thought adding a few templates to help issues and PR would benefit the project, especially having the `version` of users when they report bugs :yum:


kubouch:
Thanks, yeah, we can have these. The comments are short, they don't obscure the text fields so much :+1: .

PR #25, opened 2023-10-10
update the README

amtoine:
## Description
Nupm is evolving and, as we've already seen a few issues without even really doing any *advertising*, i think we should update the `README`.

this PR
- uses *GitHub*'s `:emojis:` to avoid fancy unicode that might not render well in editors
- mentions about `nupm` being able to install itself in a note
- talks about configuration
- talks about running tests
- adds a table of content

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,5 +45,31 @@ $env.PATH = (
 )
 ```
 
+## :test_tube: running a test suite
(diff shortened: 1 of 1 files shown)


amtoine:
i just force-pushed the last commit because i do not know how to write markdown links apparently :laughing:

kubouch:
Cool, thanks, it's looking much better!

PR #28, opened 2023-10-11
add support to install modules from non standard locations

amtoine:
## Description
- refactors a bit the tests to make them easier to work with in the long run
- adds two tests to
  1. install from `src/`: `spam_module_src_root` module and `install-module-with-src-root` test
  2. install from the root of the repo: `spam_module_repo_root` module and `install-module-with-repo-root` test
- allows to change the path to the package module files with `$.root` in the `package.nuon` file

the general idea is that
- if `$.root` is not defined, `nupm install` will try to install the module files from the subdirectory inside the root whose called the same as the package, e.g. `nu-git-manager` would install from `./nu-git-manager/`
- if `$.root` is defined, `nupm install` will copy the files from `$.root` and put that in the same place in the Nupm store

### DIFF
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -74,7 +74,7 @@ def install-path [
 
     match $package.type {
         "module" => {
-            let mod_dir = $pkg_dir | path join $package.name
+            let mod_dir = $pkg_dir | path join ($package.root? | default $package.name) | path expand
 
             if ($mod_dir | path type) != dir {
                 throw-error "invalid_module_package" (
@@ -97,7 +97,7 @@ def install-path [
                 )
             }
 
-            cp --recursive $mod_dir $module_dir
+            cp --recursive $mod_dir $destination
 
             if $package.scripts? != null {
                 log debug $"installing scripts for package ($package.name)"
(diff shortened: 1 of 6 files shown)


amtoine:
expected failure in https://github.com/nushell/nupm/actions/runs/6487491436/job/17617914677?pr=28, isn't it nice? :smirk: :pray:

amtoine:
@kubouch 
you told me we could use `$.modules: list<string>` instead of `$.root: string` but i'm still not sure how this would work :thinking: 
all the files from all the paths in `$.modules` would be installed in the same directory in the Nupm store?
because we have a single `package.nuon` file :open_mouth: 

or maybe each one of the `$.modules` should have a `package.nuon` file? :thinking:

kubouch:
The `modules` field would give a list of files / directories to be installed to `$env.NUPM_HOME | path join modules`. You could use it to change your root by setting `modules = ["your-desired-root-path/module-path"]` and not having the default module present at all. We need to be able to install multiple submodules of a package to support repos like `nu_scripts` and having both `root` and `modules` is redundant.

PR #33, opened 2023-10-15
allow installing multiple  modules from a single package

kubouch:
related to 
- https://github.com/nushell/nupm/pull/28
 
this will supersede https://github.com/nushell/nupm/pull/28

## Description
Adds support for `$.modules` field to package.nuon. Similar to `$.scripts`, this allows installing modules (both files and directories) from custom locations. 

> **Note**
> for Nupm, a package can be a collection of modules and scripts, this PR comes closer to that goal

Another change is that the default package module (`package-name/` dir) and script (`package-name.nu` file) are not required. You can have a package that has _only_ the `$.modules:` and `$.scripts:` fields in package.nuon.

### DIFF
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -38,19 +38,20 @@ def install-scripts [
 ]: list<path> -> nothing {
     each {|script|
         let src_path = $pkg_dir | path join $script
+        let dst_path = $scripts_dir | path join $script
 
         if ($src_path | path type) != file {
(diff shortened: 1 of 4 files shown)


amtoine:
> **Note**
> might supersede #28

amtoine (approved):
let's roll with this and mainly try to fix `nu_scripts` :pray: 
thanks @kubouch :partying_face:

PR #36, opened 2023-10-22
fix the logic of checking module destination

amtoine:
## Description
i'm not sure `(... | path exists) == file` is doing anything very interesting, right? :open_mouth: 

`(... | path exists)` or `(... | path type) == file` sounds more legit to me, i went with the former given the context of the command it's used in: it's checking for the existence :yum:


kubouch:
Good catch, seems like a leftover typo.

PR #37, opened 2023-10-23
refactor the CI into workflow scripts

amtoine:
## Description
this PR moves the two biggest steps of the CI to two scripts
- `set-platform-specific-variables.sh`
- `install-nightly-nushell.sh`

my idea here is to both make the CI easier to read and to allow distributing these scripts to other CIs that might need to download Nushell from the nightly builds :yum:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Set platform-specific variables
-        run: bash $CI_SCRIPTS/set-platform-specific-variables.sh "${{ runner.os }}" "$GITHUB_ENV"
+        run: bash $CI_SCRIPTS/set-platform-specific-variables.sh "${{ runner.os }}" >> "$GITHUB_ENV"
 
       - name: Install Nushell from Nightly
         run: bash $CI_SCRIPTS/install-nightly-nushell.sh "${{ env.ARCH }}" "${{ env.EXT }}" "${{ env.NU_BIN }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,9 @@ defaults:
   run:
     shell: bash
 
+env:
+  CI_SCRIPTS: "./github/workflows/scripts"
+
 jobs:
   tests:
     strategy:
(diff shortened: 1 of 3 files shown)


kubouch:
I just with someone wrote an alternative shell, so we wouldn't need to use bash scripts all the time :laughing: .

Thanks! Seems like a good idea to isolate the steps, could be useful later for something else.

fdncred:
Ya, really! Why are we not using hustcer's nushell ci action nor using nushell? 
```yaml
    - name: Setup Nushell
      uses: hustcer/setup-nu@v3.6
      with:
        version: 0.86.0
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

amtoine:
> Ya, really! Why are we not using hustcer's nushell ci action nor using nushell?
> 
this CI action would not use the latest versions of Nushell though, would it? :thinking:

fdncred:
You can see from that yaml snippet that it's using 0.86.0 in our nushell CI.

hustcer:
`setup-nu` can always use latest official release of `nushell`, I guess @amtoine want to use the latest nightly release, which is not support by `setup-nu` currently.
However, I think we can use the latest official release of `nu` to install nightly version of `nu`: just download the latest nightly version and replace the current `nu` binary (For Windows we can't replace the running binary, maybe append the nightly `nu` path to the front of `PATH` will work), then it will work afterward, or in the following steps.
BTW, it's a good idea to add nightly version support to `setup-nu`, I have already created an issue here: https://github.com/hustcer/setup-nu/issues/58

amtoine:
@hustcer @fdncred 

yeah the idea was to test Nupm against the latest versions of Nushell because, locally, everything is written and tested against the latest revision :+1: 

having that available as part of the GitHub action would be perfect, for sure :star_struck:

fdncred:
Maybe @hustcer could add because we want to dog-food nushell as much as possible and not lean on other tools unless absolutely necessary. Until that point, I guess we're stuck using bash and curl, ugh. I'm not sure it makes sense to use setup-nu to download a different version of nu. If we did decide to use setup-nu to download the nightly, we could just extract it to a known folder, that is not where setup-nu's nu is, and then call the nightly with that full path.

hustcer:
> I'm not sure it makes sense to use setup-nu to download a different version of nu. If we did decide to use setup-nu to download the nightly

I think it make sense that adding nightly version support, both [`setup-rust-toolchain`](https://github.com/actions-rust-lang/setup-rust-toolchain#inputs) and [`setup-rust-action`](https://github.com/hecrj/setup-rust-action#inputs) support `nightly` version

amtoine:
> > I'm not sure it makes sense to use setup-nu to download a different version of nu. If we did decide to use setup-nu to download the nightly
> 
> I think it make sense that adding nightly version support, both [`setup-rust-toolchain`](https://github.com/actions-rust-lang/setup-rust-toolchain#inputs) and [`setup-rust-action`](https://github.com/hecrj/setup-rust-action#inputs) support `nightly` version
>
keep us up-to-date :pray:

PR #40, opened 2023-10-27
use the Nushell GitHub action targetting nightly

amtoine:
## Description
thanks to @hustcer and the [`hustcer/setup-nu@v3.7` GitHub action](https://github.com/hustcer/setup-nu?tab=readme-ov-file#use-nu-nightly-version), we can now install the *nightly* builds of Nushell without relying on Bash scripts :pray: 

this PR implements the snippet from the doc linked above and uses `nu` as the default shell in the CI.

> **Note**
> i used this opportunity to set `NU_LOG_LEVEL` to `DEBUG` to have more information if we need that in later runs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,4 @@ jobs:
         run: version
 
       - name: Run the tests
-        run: |
-          use ./nupm/
-          nupm test
+        run: nu --commands $"use ($env.PWD)/nupm/; nupm test"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ name: continuous-integration
 
 defaults:
   run:
-    shell: bash
+    shell: nu {0}
 
 env:
   CI_SCRIPTS: ".github/workflows/scripts"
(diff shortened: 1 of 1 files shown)


amtoine:
wow i'm surprised it worked that well :open_mouth: :pray:

amtoine:
@hustcer 
i've left the scripts here for now, shall we move them to https://github.com/nushell/nightly?

the scripts are [there](https://github.com/nushell/nupm/tree/main/.github/workflows/scripts) but i think they better belong to the nightly repo rather than here if they are not used at all :yum:

hustcer:
> @hustcer i've left the scripts here for now, shall we move them to [`nushell/nightly`](https://github.com/nushell/nightly)?
> 
> the scripts are [there](https://github.com/nushell/nupm/tree/main/.github/workflows/scripts) but i think they better belong to the nightly repo rather than here if they are not used at all 😋

I think the script you add: https://github.com/nushell/nightly/blob/nightly/toolkit.nu is good enough

amtoine:
> I think the script you add: https://github.com/nushell/nightly/blob/nightly/toolkit.nu is good enough
>
okey gotcha :yum: 
to be fair, i think it should be extended to complete the full installation, i.e. extract the archive and move the files around instead of just telling the user to do that on their own, but that's something for another time :relieved: 


@kubouch 
the only question remaining for us here: **do we keep the two CI scripts?** :smirk:

kubouch:
@amtoine You mean the two .sh scripts? Preferably no. If we can get rid of them, we should :grinning: .

Let's merge this and see how it works, and then we can delete the .sh scripts.

amtoine:
> @amtoine You mean the two .sh scripts? Preferably no. If we can get rid of them, we should 😀 .
> 
> Let's merge this and see how it works, and then we can delete the .sh scripts.
>
feel free to remove them at any time if we see this goes well :yum:

PR #47, opened 2023-12-09
export internal modules manually

amtoine:
related to
- https://github.com/nushell/nushell/pull/11157

## Description
this will export the `install` and `test` modules publicly from `nupm`.


kubouch:
Let's merge it since the upstream PR is going in.

PR #57, opened 2024-01-11
spread arguments to external commands

amtoine:
related to
- https://github.com/nushell/nushell/pull/11289

## Description
this should fix the deprecation warning that we get these days


amtoine:
it's a pretty simple change, let's see how it goes.

PR #60, opened 2024-01-24
Adding basic doc generation functionality

TWSiO:
## Description

This change adds basic documentation generation functionality to nupm.

Using the current package's code, or paths to other modules it will generate a directory of markdown files based on the internal metadata about the modules and commands (data from `scope`).

The majority of the code is shamelessly stolen from @amtoine's [nu-git-manager toolkit script](https://github.com/amtoine/nu-git-manager/blob/1f4153da8db3b59eafe0fac86fadb84cca2e6e15/toolkit.nu) and modified.

There's a lot of basic functionality missing that I'd like to add like generating HTML and documenting plugins, but this is a good first step.

I also added the generated documentation for nupm itself because it seemed appropriate.

### DIFF
--- a/docs/commands/README.md
+++ b/docs/commands/README.md
@@ -1,0 +1,8 @@
+## Modules
+- [`nupm`](./nupm/README.md)
+
+## Commands
+- [`nupm`](nupm/nupm.md)
+- [`doc`](doc/nupm/doc.md)
+- [`install`](install/nupm/install.md)
+- [`test`](test/nupm/test.md)
--- a/docs/commands/nupm/README.md
+++ b/docs/commands/nupm/README.md
@@ -1,0 +1,11 @@
+# Module `nupm`
+## Description
+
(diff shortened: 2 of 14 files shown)


kubouch:
Before adding `nupm doc` and committing to this particular style, we should first have a discussion about how a doc autogeneration would work in Nushell in general. _Then_ add support for it in `nupm`.  So far, we haven't discussed this thoroughly in the core team. I'm not saying this is a bad doc style, and we might as well use it as a starting point for the discussion, but I don't feel comfortable committing it to nupm _yet_ because it might as well change significantly (e.g., we could add this functionality to Nushell itself, or use a separate program to do the doc autogeneration). One approach would be to create a `nupm experimental` subcommand group and move the `doc` there without commitment, just to test it out.

amtoine:
marking as DRAFT for now

PR #64, opened 2024-01-28
Add toolkit

kubouch:
## Description

Adds a simple toolkit module that sets a temporary NUPM_xxx environment variables for development to avoid messing up your own setup.

### DIFF
--- a/toolkit.nu
+++ b/toolkit.nu
@@ -1,0 +1,9 @@
+export-env {
+    $env.NUPM_HOME = ($env.FILE_PWD | path join _nupm_dev)
+    $env.NUPM_CACHE = ($env.FILE_PWD | path join _nupm_dev cache)
+    $env.NUPM_TEMP = ($env.FILE_PWD | path join _nupm_dev tmp)
+
+    print $'NUPM_HOME:  ($env.NUPM_HOME)'
+    print $'NUPM_CACHE: ($env.NUPM_CACHE)'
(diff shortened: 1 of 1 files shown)


kubouch:
@amtoine I'll merge it since I need to use it now, but please protest if you don't like it :smile:.

amtoine:
mm, i'm not a big fan of it :laughing: 

two main reasons
- because i use an automatic _toolkit-loading_ hook i prefer the way i do it with NGM, i.e. use a [dedicated command to run the application in a sandbox](https://github.com/amtoine/nu-git-manager/blob/main/toolkit.nu#L69), on demand only
- the paths are broken on my system, might be due to the fact i use Nushell to run my window manager, e.g. `NUPM_HOME` is `/home/amtoine/.config/bspwm/_nupm_dev` :thinking:

amtoine:
but apart from that, the idea is interesting :+1:

kubouch:
OK, so how about that it wouldn't export the env inside `export-env` but it would give you a `def --env set-dev-paths` or something that you can call that would set those env vars?

amtoine:
i think it would be better :+1: 

because, right now, i can't run the following anymore :cry: 
```nushell
use nupm
nupm install --path --force .
```
to install Nupm itself, it tries to put it in the dev environment :open_mouth:

kubouch:
OK, I'll fix it up.

PR #67, opened 2024-01-28
Move toolkit environment setting to a command

kubouch:
## Description

Instead of setting the environment on module import, it is now done in a command that needs to be called manually.

### DIFF
--- a/toolkit.nu
+++ b/toolkit.nu
@@ -1,11 +1,15 @@
-export-env {
-    $env.NUPM_HOME = ($env.FILE_PWD | path join _nupm_dev)
-    $env.NUPM_CACHE = ($env.FILE_PWD | path join _nupm_dev cache)
-    $env.NUPM_TEMP = ($env.FILE_PWD | path join _nupm_dev tmp)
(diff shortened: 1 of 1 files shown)


toolkit.nu
```diff
@@ -1,11 +1,15 @@
...
 
-    $env.PATH ++= [($env.FILE_PWD | path join _nupm_dev scripts)]
-    $env.NU_LIB_DIRS ++= [($env.FILE_PWD | path join _nupm_dev modules)]
+    $env.PATH ++= [('./_nupm_dev/scripts' | path expand)]
+    $env.NU_LIB_DIRS ++= [('./_nupm_dev/modules' | path expand)]
 
+    print-nupm-env
+}
+
+export def print-nupm-env [] {
```
amtoine, line 12: in order for this command not to give an error, i would use `$env.NUPM_HOME?` and same with the other ones, e.g. i don't have `$env.NUPM_CACHE` defined and calling `print-nupm-env` crashes :wink:

amtoine (approved):
thanks for the quick update :pray: 

let's land this :+1:

PR #56, opened 2023-12-28
Allow online installation of packages

kubouch:
## Registries

_See the newly added registry.nuon file for more specific info about the registry format._

This PR adds a new concept: **registry**. Registries are .nuon files containing a list of packages, their versions, and how they should be installed. Currently, registry can contain two kinds of packages:
* `local` -- installed from local file system
* `git` -- installed from remote Git repository

The list of recognized registries is stored in `$env.NUPM_REGISTRIES` as a record of name: path pairs. That way, users can maintain their collection of sources from where they want to install packages. Not sure if this will be the final way to pass available registries to nupm, but should be good enough for early testing.

## Online Installation

If a package is to be installed from a Git repository, the repository is cached in `$env.NUPM_CACHE` to avoid re-cloning the same repository all the time. Once cloned, the local path of the cloned repository is passed to the installer, as if you called `nupm install --path=...`.

One important missing feature is some kind of file integrity check. One step at a time...

## Versions

If a registry contains multiple versions (common scenario), by default, the newest one is installed. "Newest" currently means sorting the versions alphabetically and taking the last one. If `--pkg-version` is specified, nupm tries to install the exact version passed to this option. This needs better version matching (e.g., version 0.2 matching 0.2.1 etc.), but again, one step at a time...

## New command

`nupm search` searches for packages in registries without installing them. I thought this would be useful to have early on.

## Try it out

From the repository root:

```nushell
overlay use nupm/
nupm install nu-git-manager
# use nu-git-manager *
# ...
```

## TODO

- [x] Remove debug stuff
- [x] Nicer / more informative prints
- [x] (can be done over time) Rewrite https://github.com/nushell/nupm/blob/main/index.nuon to fit the registry format, move it to its own repo
	- It would be a single-file repository
	- Package authors would put PRs there whenever they update their package
- [x] Tests

## After
- better version matching => `0.2` shouldn't match `0.2.1`
- file integrity checks

PR #69, opened 2024-02-18
update registry by adding all known packages

amtoine:
followup to #56 

## Description
this should update the `registry.nuon` file thanks to the content of previous `index.nuon` (see cbca144 for the pipeline used to update the registry).

this PR uses the latest version of each package.

## try it out
```nushell
nupm search --registry ./registry.nuon explore
    | into record
    | update pkgs { into record }
    | table --expand
```
```
────┬──────────────────────────────────────────────────────
name│registry
path│./registry.nuon
    │────────┬─────────────────────────────────────────────
pkgs│name    │nu_plugin_explore
    │version │0.1.2
    │url     │https://github.com/:amtoine/nu_plugin_explore
    │revision│0.1.2
    │path    │.
    │type    │git
    │────────┴─────────────────────────────────────────────
────┴──────────────────────────────────────────────────────
```


registry.nuon
```diff
@@ -26,21 +26,166 @@
...
+            version,
+            url,
+            revision,
+            path
+        ];
+        [
+            nu_plugin_audio_hook,
+            "0.1.12",
+            "https://github.com/FMotalleb/nu_plugin_audio_hook",
+            "0.1.12",
```
kubouch, line 40: I think the version and revision are a bit mixed on this one. The last version and revision seems to be "v0.1.2".

registry.nuon
```diff
@@ -26,21 +26,166 @@
...
+            .
```
kubouch, line 41: I feel like we should allow `null` here.

amtoine: yeah, could allow that as well :+1:

registry.nuon
amtoine: i did not look at these other packages, as i do not maintain them.
i can have a look real quick

kubouch (approved):

PR #76, opened 2024-03-08
add `main` branch of `tmux-sessionizer`

amtoine:
## Description
[`tmux-sessionizer`](https://github.com/amtoine/tmux-sessionizer) does not really have versions because i've been lazy with that package of mine :rofl: 

i thought adding the `main` branch as a version could be nice.


kubouch:
It's kind of abusing how the versioning works currently, but good enough for now :-D

kubouch (approved):

PR #75, opened 2024-03-08
:bug: fix extra `)`

AucaCoyan:
I've compiled nushell from latest `main` commit and threw something like:
```
Error: nu::parser::unbalanced_delimiter

  × Unbalanced delimiter.
    ╭─[C:\Users\AucaMaillo\other-repos\nupm\nupm\utils\misc.nu:23:13]
 22 │                 + $" ($missing_cols | str join ', ')")
 23 │             )
    ·             ┬
    ·             ╰── unbalanced ( and )
 24 │         }
    ╰────
```
So I corrected the typo.

PR #77, opened 2024-03-08
Fix unclosed paren

kubouch:
## Description

There was apparently uncaught syntax error that got fixed in a recent Nushell PR


fdncred:
Sounds similar to https://github.com/nushell/nupm/pull/75

PR #78, opened 2024-03-08
Add `nupm status`; List all files of module

kubouch:
## Description

I implemented a way how to fetch all files used by a `module` package (by calling `use module; view files ...`). This will later allow generating a package's file list, which will be used for file integrity checks.

Implemented a `nupm status` as a proof-of-concept to demonstrate this.

Currently, `script` packages are unsupported because it would require calling `source` => it would evaluate the package which we don't want. I'm considering updating Nushell's `nu-check` and making it optionally list all the parsed files.

`custom` packages do not list any files. Since they don't have any known structure, package authors would probably need to put a list of installed files to `nupm.nuon` (same will be for plugins).

`$.scripts` fiels in `nupm.nuon` is respected.

### DIFF
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -4,39 +4,9 @@ use utils/completions.nu complete-registries
 use utils/dirs.nu [ nupm-home-prompt cache-dir module-dir script-dir tmp-dir ]
 use utils/log.nu throw-error
 use utils/misc.nu check-cols
+use utils/package.nu open-package-file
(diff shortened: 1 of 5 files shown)


amtoine (approved):
could try it and it looks fine to me :ok_hand: 

looking forward to the rest :slightly_smiling_face:

kubouch:
The files listing will need a parser support, but we can still land this, it should work fine with basic modules.

PR #81, opened 2024-03-15
rename default registry

amtoine:
## Description
i thought it was a bit strange to read `nupm_test` in the output of `nupm search` for the official registry of Nushell :open_mouth: 

this PR renames it to `nushell` for now.
maybe `official`?


kubouch:
Well, it's not an official Nushell registry, it's just a proof of concept for testing nupm. I'd avoid having anything “official” until we have it more fleshed out.

PR #84, opened 2024-03-24
add `nu-fonts-install` and `nu-discord-update`

amtoine:
## Description
this PR adds the two following packages, taken from my config, to the registry:
- [`nu-fonts-install`](https://github.com/amtoine/nu-fonts-install): install fonts easily, supports _Nerd_ fonts and the GitHub's Monaspace ones
- [`nu-discord-update`](https://github.com/amtoine/nu-discord-update): updates the Discord client when it complains about updates on Arch at least lol

### DIFF
--- a/registry.nuon
+++ b/registry.nuon
@@ -305,6 +305,20 @@
             "https://github.com/amtoine/tmux-sessionizer",
             "main",
             .
+        ],
+        [
+            nu-discord-update,
+            "0.1.0",
+            "https://github.com/amtoine/nu-discord-update",
+            "0.1.0",
+            .
+        ],
+        [
+            nu-fonts-install,
+            "0.1.0",
+            "https://github.com/amtoine/nu-fonts-install",
+            "0.1.0",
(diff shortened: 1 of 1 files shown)


fdncred:
related/unrelated - seems like the nupm.nuon file should say what platform scripts are supported on. For instance, the nu-discord-update wouldn't work on mac or windows and nu-fonts-install might work on mac but not windows. Not to mention, they both use externals which could be problematic if they don't exist. Just wondering how that type of stuff is handled or planned to be handled in the future?

amtoine:
very good points @fdncred :+1:

kubouch:
Good idea about the platform support.

We plan to have dependencies eventually where packages can depend on other nupm packages. You could package an external as a nupm package (we need to support binary packages anyway because of plugins). A second option could be adding "external dependencies" that would be checked via `which` and if not found, an error will be printed that you need to have them installed.

kubouch (approved):

PR #86, opened 2024-03-26
Redesign registry; Allow publishing to registry

kubouch:
## Description

This changes registry to from *one* file that is *manually updated* to *multiple* files that are *auto-generated*. Now, the registry should be only updated with the new `nupm publish` command.

See the [doc page](https://github.com/nushell/nupm/blob/bde509cfd9a53939c7a4df01564136e1f9c0b2ee/docs/design/registry.md) for more details and the `registry/` directory for the new proposed structure.

### Misc changes

* The structure of the registry is also different. The git/local package type is now encoded in a flat table instead of being a separate record. See `registry/` and `tests/packages/registry/` for examples.
* The `path` field can be `null`, not requiring the "."

### TODO

- [x] Add package file hashing to stop redownloading them all the time


registry/registry.nuon
amtoine, line 1: a comment for all the registry NUON files: would it be better to have them not on a single line? i guess having them in a pretty format, e.g. with `to nuon -i 4` would allow easier debugging and seeing what changed from one revision to another?

i mean, this is not critical, it's always possible to keep that compact format and use `open`, `to nuon -i` and things like `diff -u` to see the "_real_" diff between the NUON data :wink:

toolkit.nu
```diff
@@ -13,6 +27,13 @@ export def print-nupm-env [] {
...
     print $'NUPM_CACHE: ($env.NUPM_CACHE?)'
     print $'NUPM_TEMP:  ($env.NUPM_TEMP?)'
-    print $'PATH: ($env.PATH?)'
+    if $nu.os-info.family == 'windows' and 'Path' in $env {
+        print $'Path: ($env.Path?)'
+    } else if 'PATH' in $env {
+        print $'PATH: ($env.PATH?)'
+    } else {
+        print 'no PATH env var'
+    }
```
amtoine, line 36: might be able to write a simpler
```nushell
print $"PATH: ($env.PATH? | default $env.Path? | default [])"
```
but this `if` is fine, as you want :+1:

tests/mod.nu
```diff
@@ -147,3 +147,28 @@ export def env-vars-are-set [] {
...
+        let tmp_reg_file = [
+            $env.NUPM_TEMP packages registry test_registry.nuon
+        ]
+        | path join
+
+        touch $tmp_reg_file
+
+        [spam_script spam_script_old spam_custom spam_module] | each {|pkg|
+            cd ([tests packages $pkg] | path join)
+            nupm publish $tmp_reg_file --local --save --path $'../($pkg)'
```
amtoine, line 166: would the following be more robust on Windows? (yeah i'm always worried about Windows lol)
```suggestion
            nupm publish $tmp_reg_file --local --save --path (".." | path join $pkg)
```

tests/mod.nu
```diff
@@ -147,3 +147,28 @@ export def env-vars-are-set [] {
     assert equal $env.NUPM_CACHE $dirs.DEFAULT_NUPM_CACHE
     assert equal $env.NUPM_REGISTRIES $dirs.DEFAULT_NUPM_REGISTRIES
 }
+
+export def generate-local-registry [] {
+    with-test-env {
+        # cp -r tests/packages $env.NUPM_TEMP
```
amtoine, line 153:
```suggestion
```

amtoine (changes requested):
just a few remarks but, otherwise, all looks pretty cool.
it's not easy to test right now because the new registry URL is invalid until we land that PR...



## the registry files
i've been able to make sure all the registry files were valid, which is very nice.
below is the script that passes :ok_hand:
```nushell
use std assert

def make-it-similar []: table<name: string, version: string, url: string, revision: string, path: string> -> table<name: string, version: string, url: string, revision: string, path: string> {
    sort-by version name | select name version url revision path
}

let old = http get https://raw.githubusercontent.com/nushell/nupm/main/registry.nuon
    | get git
    | make-it-similar

let new = open registry/registry.nuon
    | get path
    | each {|it| open ("registry/" | path join $it)} # open each registry package file
    | flatten                  # extract the path and revision from `$.info`
    | update info { flatten }  # |
    | flatten --all            # |
    | reject type
    | default "." path         # remove the `null`
    | make-it-similar

assert equal $new $old
print "all good"
```

amtoine:
windows being painful again...

registry/registry.nuon
kubouch: It takes way too much space with `-i 4`, but being able to diff properly is nice... You can also use a better diff tool like difftastic, but it's still quite hard to see.

I'm kind of leaning towards keeping it as-is because it's meant to be processed by machines only. Maybe we can roll with it for a while and see if it becomes a pain to debug and switch to `-i 4` if so?

kubouch:
The hash is failing. I'm wondering if it's Windows vs. Unix newlines.

registry/registry.nuon
amtoine: yup, we can definitely do that!

kubouch:
Related to diffing the nuon files, inspired by https://www.youtube.com/watch?v=Md44rcw13k4. Pasting it here from Discord so it won't get burried:
* create a new file called .gitattributes
* put the following like in it `*.nuon diff=nuon`
* run the following command `git config diff.nuon.textconv nu`
* modify a NUON file and see how the diff is pretty

kubouch:
@amtoine I fixed the Windows issue! It wasn't newlines but the --path flag which was constructed with `\` on Windows and `/` on Mac/Linux. I forced it to always have `/`. Feel free to approve and merge!

kubouch:
One argument for keeping the nuon files "raw" is that they do not contain newlines, so their hashes should be the same on Windows/non-Windows. If we made the nuon files more human-friendly, they would contain newlines which could potentially mess things up: Simply opening and closing a nuon file could change the line endings in the entire file, making its hash not match the expected hash. Maybe it would be fine, but we'd have to make sure and test it, I'd rather avoid this by having no newlines, at least for now.

texastoland:
Just devil's advocate but could you manually replace all `\r` with empty string before saving?

kubouch:
> Just devil's advocate but could you manually replace all `\r` with empty string before saving?

Yes, that's an option (or convert all `\n` to `\r\n`). I chose to side step it entirely for now. I'm a bit afraid that once we start messing with newlines, we'd open a can of works. Like I mentioned above, I can imagine just viewing the file in some smart editor could convert the line endings automatically. Then there is also Git which can convert the line endings depending on user's preferences. It's not something I want to mess with right now 😆 .

amtoine:
nice job @kubouch !!

Windows is such a pain...
in `nu-git-manager`, i wrote a `path sanitize` command that makes Windows path better (lol) and i have to use it almost on all paths :rofl:

nupm/publish.nu
```diff
@@ -0,0 +1,241 @@
...
+            }
+        }
+        'local' => {
+            if $info != null {
+                throw-error "Package type 'local' must have null info."
+            }
+        }
+    }
+
+    let path = if $path != null { $path | str replace '\' '/' }
```
amtoine, line 110: shouldn't all `\` be replaced by `/`?
```suggestion
    let path = if $path != null { $path | str replace --all '\' '/' }
```

amtoine (changes requested):
the new commits look fine to me :ok_hand: 

there's just a review question and some missing packages :wink: 
```nushell
use std assert

def make-it-similar []: table<name: string, version: string, url: string, revision: string, path: string> -> table<name: string, version: string, url: string, revision: string, path: string> {
    sort-by version name | select name version url revision path
}

let old = http get https://raw.githubusercontent.com/nushell/nupm/main/registry.nuon
    | get git
    | make-it-similar

let new = open registry/registry.nuon
    | get path
    | each {|it| open ("registry/" | path join $it)} # open each registry package file
    | flatten                  # extract the path and revision from `$.info`
    | update info { flatten }  # |
    | flatten --all            # |
    | reject type
    | default "." path         # remove the `null`
    | make-it-similar

if $new != $old {
    $new | to nuon --indent 4 | save --force /tmp/new.nuon
    $old | to nuon --indent 4 | save --force /tmp/old.nuon
    diff -u /tmp/old.nuon /tmp/new.nuon
}
print "all good"
```
gives me
```diff
--- /tmp/old.nuon       2024-05-05 13:15:41.093046938 +0200
+++ /tmp/new.nuon       2024-05-05 13:15:41.093046938 +0200
@@ -245,55 +245,6 @@
         .
     ],
     [
-        nu_plugin_explore,
-        "0.92.0",
-        "https://github.com/amtoine/nu_plugin_explore",
-        "0.92.",
-        .
-    ],
-    [
-        nu_plugin_clipboard,
-        "0.92.1",
-        "https://github.com/FMotalleb/nu_plugin_clipboard",
-        "98d7102",
-        .
-    ],
-    [
-        nu_plugin_explore,
-        "0.92.3-2595f31541554c6d8602ebebc9dffbc4dd29dd89",
-        "https://github.com/amtoine/nu_plugin_explore",
-        "0.92.3-2595f31541554c6d8602ebebc9dffbc4dd29dd89",
-        .
-    ],
-    [
-        nu_plugin_explore,
-        "0.92.3-55edef5ddaf3d3d55290863446c2dd50c012e9bc",
-        "https://github.com/amtoine/nu_plugin_explore",
-        "0.92.3-55edef5ddaf3d3d55290863446c2dd50c012e9bc",
-        .
-    ],
-    [
-        nu_plugin_explore,
-        "0.92.3-be5ed3290cb116cbe9f3038f7a2d46aaac85a25c",
-        "https://github.com/amtoine/nu_plugin_explore",
-        "0.92.3-be5ed3290cb116cbe9f3038f7a2d46aaac85a25c",
-        .
-    ],
-    [
-        nu_plugin_explore,
-        "0.92.3-c06ef201b72b3cbe901820417106b7d65c6f01e1",
-        "https://github.com/amtoine/nu_plugin_explore",
-        "0.92.3-c06ef201b72b3cbe901820417106b7d65c6f01e1",
-        .
-    ],
-    [
-        nu_plugin_explore,
-        "0.93.0",
-        "https://github.com/amtoine/nu_plugin_explore",
-        "0.93.0",
-        .
-    ],
-    [
         nu_plugin_port_scan,
         "1.0.10",
         "https://github.com/FMotalleb/nu_plugin_port_scan",
```
related to
- #87 
- #88

nupm/publish.nu
kubouch: Yes, you're right, I forgot about that.

kubouch:
Ah yeah, it's missing the added packages. I currently don't have time to add them. If you have time, you can merge it and publish them again to the registry using the new `nupm publish`. This way it would get tested by someone apart from me. Otherwise, I can do it whenever I find the time.

kubouch:
Now thinking about it, I think we should support only semver-style versions because we need to be able to sort by version and select the latest. The same with the `main` version. At least in the official registry, having random strings would be too confusing. I understand the value of having, e.g., the `main` as an always up-to-date package, but it really messes up the version resolution.

I'd propose keeping only semver in the official registry, but let's keep experimenting with being able to have some alternative versions.

texastoland:
> I think we should support only semver-style versions

Commit hashes are [supported in semver](https://semver.org/#spec-item-10) (in case it helps).

kubouch:
TIL, but I still think we need to restrict our versions to a subset that doesn't prevent sorting. Otherwise, we wouldn't be able to reliably update a package. You want to type `nupm update nu_plugin_explore` and get the latest version. Now, which one of the four different `0.92.3`s is the newest?

We could support having a `null`/`main`/`latest` in place of the version. In that case, this version would always be considered “newest” and would be always re-fetched when you call `nupm update`. `nupm install` would also always install this version if you don't specify the version manually. This would be useful for cases like the current `tmux-sessionizer` tracking the main branch. That would be OK for a private / unofficial registry. For the official registry, we'd disallow it.

texastoland:
> I still think we need to restrict our versions to a subset that doesn't prevent sorting.

I trust your vision for it ✌🏼 FWIW the spec is `major.minor.patch-alphabetical+ignored`.

amtoine:
i agree @kubouch, maybe these deps are simply ill-defined :eyes: 

does that mean that
- the two i cannot add shouldn't be added?
- the ones that are not proper semver should be removed?

then we can land this PR :pray:

kubouch:
@amtoine I republished the problematic packages, should be good now. The problem with nu_plugin_explore is that revisions up to 0.93.0 still had the old version 0.1.2 in their nupm.nu, so they're not publishable because 0.1.2 is already published. Seems like you forgot to update it.

amtoine:
aaah yeah, you're right :sob: 

we're good to land then? i kinda lost track of whole the changes lol

amtoine:
still think the registry diffs are impossible to review, which is a bit of a bummer imo :confused:

texastoland:
Same. I think there'd be some value in standardizing on Unix line endings. Otherwise DX will be a step above binary. Epic regardless 👏🏼

kubouch:
Thinking about it, I think what we can do is to make `nupm publish` generate one type of line ending (LF or CRLF), but more importantly, we can change the hash function to replace all CRLF with LF in text files and hash that. That should give us identical hashes on Windows and Unix.

I'm going to merge this as-is, but I'll play with the newlines in the next PR, or anyone else can take a shot at it. I think relying forever on having zero newlines doesn't sound very robust in the long run.

PR #103, opened 2024-10-18
Import std/log environment; CI: Update Ubuntu

FMotalleb:
Will fix [this issue](https://github.com/nushell/nushell/issues/14109) but its not a good fix anyway 

## Description

This is a hotfix PR to address an unusual issue with the `std log` package in Nushell. When the log is imported during the initialization stage, it doesn’t persist long enough to be fully available at runtime, specifically, the environment variables seem to disappear, although the log function itself remains accessible. This is a temporary fix and might not be the final solution. It could potentially be a bug in Nushell, but unfortunately, I don't have the time to investigate it further.

for more information please check [this issue](https://github.com/nushell/nushell/issues/14109)

I think this fix changes the output of the nupm, Plz close this PR if you found any real solution

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        platform: [windows-latest, macos-latest, ubuntu-20.04]
+        platform: [windows-latest, macos-latest, ubuntu-24.04]
 
     runs-on: ${{ matrix.platform }}
 
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -238,6 +238,7 @@ export def main [
     --force(-f)  # Overwrite already installed package
     --no-confirm  # Allows to bypass the interactive confirmation, useful for scripting
 ]: nothing -> nothing {
+    use std/log
     if not (nupm-home-prompt --no-confirm=$no_confirm) {
         return
     }
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -1,5 +1,3 @@
-use std log
-
 use utils/completions.nu complete-registries
 use utils/dirs.nu [ nupm-home-prompt cache-dir module-dir script-dir tmp-dir ]
 use utils/log.nu throw-error
@@ -238,7 +236,6 @@ export def main [
     --force(-f)  # Overwrite already installed package
     --no-confirm  # Allows to bypass the interactive confirmation, useful for scripting
 ]: nothing -> nothing {
-    use std/log
     if not (nupm-home-prompt --no-confirm=$no_confirm) {
         return
     }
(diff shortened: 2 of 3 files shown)


amtoine:
why only add this to `nupm install` and not the other commands?

FMotalleb:
I didn't knew where to put it, actually, now if you check the CI its broken for the same reason
https://github.com/nushell/nupm/actions/runs/11405030607/job/31741019156?pr=103

amtoine:
mm that looks quite annoying :sweat_smile: 

sounds like something a library shouldn't have to worry about, right?

fdncred:
@NotTheDr01ds thoughts on how to get rid of this error?

FMotalleb:
well after some investigation I found out that when you have a module like this

example.nu
```nu
use std/log

export def test_cmd [param: string, ] {
  log info $param
}
```
then running `use example.nu`

the imported function is working but the env vars aren't

```
~/.../test> example test_cmd test                                                                                                         10/19/2024 07:14:36 PM PM
Error: nu::shell::column_not_found

  × Cannot find column 'NU_LOG_FORMAT'
     ╭─[std/log\mod.nu:158:9]
 157 │     let log_format = if ($format_string | is-empty) {
 158 │         $env.NU_LOG_FORMAT
     ·         ─────────┬────────┬
     ·                  │        ╰── value originates here
     ·                  ╰── cannot find column 'NU_LOG_FORMAT'
 159 │     } else {
     ╰────
```

NotTheDr01ds:
Apologies - missed the notification again.

This goes back to https://github.com/nushell/nushell/issues/13403.  I'm not sure if it's a bug or design, but importing a module *within* another module doesn't import its `export-env` block.

The workaround is to make sure the parent module re-exports the child module's (`std/log` here) environment:

```nushell
use std/log

export-env {
  use std/log []
}

export def test_cmd [param: string, ] {
  log info $param
}
```

NotTheDr01ds:
I thought about defining these at startup and just taking them out of `std/log`, but this is going to be a problem every time we add new exported environment variables to a module (`std` or not). 

Technically, it's also a problem for the `dirs` module, but since that provides *user* functionality, it's not something that's likely to get imported in another module.

It would be great if the root issue could be resolved, but perhaps we workaround it for now by simply exporting them in startup?

kubouch:
If I understand correctly, the problem is that environment variables from `std/log` are not available right? Then, I think the `export-env` solution is the best bet. It works like that "by design", but it's not a very good design, I'll elaborate more in the linked issue. For this PR, I'd propose go with the `export-env` method.

FMotalleb:
well "It Doesn't work on my machine" 😂
I replaced `use std log` in install subcommand (the only location that std log is imported) with
```
use std/log

export-env {
    use std/log []
}
```

I still face the same error (`cannot find column 'NU_LOG_FORMAT'`)

FMotalleb:
oh my bad there was another export-env after my block

kubouch:
OK, looks good, thanks!

PR #107, opened 2024-11-21
Fix: Hash detection affects all Search results

mrxiaozhuox:
## Description

The current Hash check in the code will directly cause the program to throw an error. When using Search, if there is even one Hash check failure, no information can be displayed.

I have fixed this issue and retained the error throw in the Install section. However, I don't think it's necessary to make the program fail to run specifically because of this issue.

The possible useful solutions I have thought of, which I can modify if you are willing to accept, are:

1. Automatically ignore all packages with Hash check failures.
2. Give users the right to choose, informing them that the content does not match, but they can still install it.

### DIFF
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -207,6 +207,11 @@ def fetch-package [
         throw-error $'No package matching version `($version)`'
     }
 
+    if $pkg.hash_mismatch != true {
+      throw-error ($'Content of package file ($pkg.path)'
+                        + $' does not match expected hash') 
+    }
+
     print $pkg
 
     if $pkg.type == 'git' {
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -207,9 +207,9 @@ def fetch-package [
         throw-error $'No package matching version `($version)`'
     }
 
-    if $pkg.hash_mismatch != true {
+    if $pkg.hash_mismatch == true {
       throw-error ($'Content of package file ($pkg.path)'
-                        + $' does not match expected hash') 
+                        + $' does not match expected hash')
     }
 
     print $pkg
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -207,6 +207,11 @@ def fetch-package [
         throw-error $'No package matching version `($version)`'
     }
 
+    if $pkg.dirty {
+        throw-error ($'Content of package file `($pkg.path)'
+            + $'` does not match expected hash.')
+    }
+
     print $pkg
 
     if $pkg.type == 'git' {
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -207,11 +207,6 @@ def fetch-package [
         throw-error $'No package matching version `($version)`'
     }
 
-    if $pkg.dirty {
-        throw-error ($'Content of package file `($pkg.path)'
-            + $'` does not match expected hash.')
-    }
-
     print $pkg
 
     if $pkg.type == 'git' {
--- a/nupm/search.nu
+++ b/nupm/search.nu
@@ -10,7 +10,7 @@ export def main [
     --pkg-version(-v): string  # Package version to install
     --exact-match(-e)  # Match package name exactly
 ]: nothing -> table {
-    search-package $package --registry $registry --exact-match=$exact_match --skip-dirty-pkg
+    let result = search-package $package --registry $registry --exact-match=$exact_match
     | flatten
     | each {|row|
         {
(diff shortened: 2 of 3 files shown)


mrxiaozhuox:
It seems that directly adding a `dirty` field in the `pkgs` is not a good idea, so we can only wrap a layer of status or directly filter out the data that failed the hash verification.

mrxiaozhuox:
or add a new flag to decide how to control hash check failed.

mrxiaozhuox:
I pushed a new update to fix the issue without affecting the structure.

amtoine:
hopefully superseded by https://github.com/nushell/nupm/pull/111

mrxiaozhuox:
> hopefully superseded by #111

I believe there is a substantial difference between these two PRs. Your PR seems to fix the Hash operation, but what this PR does is: if there is a Hash check failure, it does not affect the results of `nupm search`, because a hash check failure would directly terminate the entire program, and if there is one failure in the registry, none of the search results can be displayed.

Do you intend to fix this issue in your PR?

Thank you.

amtoine:
aah okok my bad.

well yeah, maybe once the fix lands, we can think about what should happen if a hash fails :wink:

kubouch:
Thanks for the PR, sorry for the late reply.

I think the logic could be simplified: Inside `search-package`, don't throw an error at all, but just add a new column “hash_mismatch” (instead of “dirty”, I think it's more descriptive) with true/false depending on whether the hash failed or not. Then, applications that want to throw an error (`nupm publish`, `nupm install`) can do so afterward, and `nupm search` can just filter it out. `nupm search` could return the search result, but if there are some mismatches, it could print a warning, for example.

kubouch:
Btw, you might experience some type-related errors due to the last Nushell update. I fixed those on the main branch.

mrxiaozhuox:
> Again, sorry for the late reply, I don't have much time to consistently work on nupm.
> 
> There are still test failures. There are some hash mismatches and `nupm publish` seems to be looking for the `hash-mismatch` column in the test registry package file of `spam_script`: https://github.com/nushell/nupm/blob/main/tests/packages/registry/spam_script.nuon . So I think these need to be addressed, but otherwise the PR seems good.

do I need edit tests/nu.nu or just https://github.com/nushell/nupm/blob/main/tests/packages/registry/spam_script.nuon this file?

kubouch:
I haven't looked closely, but the `hash-mismatch` shouldn't be a part of the registry package files (like the `spam_script.nuon`). So `nupm publish` (and other relevant commands) should be modified to not require this column in those files.

mrxiaozhuox:
Only this part used search-package, but only take `type` field, so I don't know why test will check `hash-mismatch`:
https://github.com/nushell/nupm/blob/e9f39eac45f8043261f18e85080f487a757f7d95/nupm/publish.nu#L50-L61

Maybe because this? 
https://github.com/nushell/nupm/blob/335a5775ab0a36572a619e1f110938c5f03db7ae/nupm/utils/registry.nu#L11
did test will check this var?

mrxiaozhuox:
> I haven't looked closely, but the `hash-mismatch` shouldn't be a part of the registry package files (like the `spam_script.nuon`). So `nupm publish` (and other relevant commands) should be modified to not require this column in those files.

I found a cols check at here:
https://github.com/nushell/nupm/blob/e9f39eac45f8043261f18e85080f487a757f7d95/nupm/publish.nu#L234-L240

I want delete `hash-mismatch` at `REG_PKG_COLS`, because I think this only a temp field. how do u think?

kubouch:
> I want delete `hash-mismatch` at `REG_PKG_COLS`, because I think this only a temp field. how do u think?

Yeah, this sounds reasonable.

mrxiaozhuox:
> > I want delete `hash-mismatch` at `REG_PKG_COLS`, because I think this only a temp field. how do u think?
> 
> Yeah, this sounds reasonable.

I deleted `hash_mismatch` field.

kubouch:
Good, there are still some test failures, can you fix those?

mrxiaozhuox:
> Good, there are still some test failures, can you fix those?

sorry for late, I don't know how to fix this error, look's like test package and registry's hash not match.

kubouch:
You can update the hash in the registry to match the package hash.

mrxiaozhuox:
> You can update the hash in the registry to match the package hash.

OMG, I fixed it.

I checked the MD5-related content many times and finally found that it was actually a conditional statement error. 😢

kubouch:
That happens :D Thanks!

PR #111, opened 2024-11-25
fix hash computations

amtoine:
with Nushell 0.100.0, `to nuon` has changed and now quotes strings that contain a `.`, thus breaking hash computations...

this PR does the following
1. formats the registry NUON file so that it's easier to read and maintain
2. relax the "newline" test on registry files, not sure why we _want_ to have newlines in registry files
3. fix the computation of the hash of pkg files by using `to nuon` instead of `open --raw`
4. pin the Nushell version to 0.100.0 in the CI because having a moving target in a CI is very hard to work with

> **Note**
>
> point 1, 2 and 3 are probably debatable and not mandatory for this PR but it was by far the easiest to get this PR going :yum:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
 
       - uses: hustcer/setup-nu@v3.10
         with:
-          version: nightly
+          version: "0.100.0"
 
       - name: Show Nushell Version
         run: version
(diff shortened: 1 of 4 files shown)


kubouch:
The newline test was me working around Windows \r\n vs. Unix \n. I guess the nuon route is probably better. Since the tests now pass, let's ship it! Thanks!

PR #112, opened 2024-12-22
Fix parsing errors for nushell nightly

Bahex:
## Description

> Fixes the invalid type signature that is now caught by [nushell/nushell#14510](https://github.com/nushell/nushell/pull/14510)


fdncred:
Thanks

PR #114, opened 2025-01-24
docs: added basic usage instructions

vyadh:
## Description
Getting started with Nupm's current basic functionality required reverse-engineering the needed commands from the code, so I thought a good start would be to define this in the readme.

Some of the existing detailed commands weren't quite accurate so I've also fixed those.

Note that I'm only documenting the functionality I've used for now. It probably makes sense to add the registry stuff once it's more fleshed out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,9 @@ $env.PATH = (
 
 ## :rocket: usage [[toc](#table-of-content)]
 
-The following uses a fictional `foo` package for examples.
+Nupm can install different types of packages, such as modules and scripts. It also provides a mechanism for a custom installation using a `build.nu` file.
+
+As an illustrative example, the following demonstrates use of a fictional `foo` module-based package.
 
 ### install a package [[toc](#table-of-content)]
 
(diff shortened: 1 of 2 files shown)


kubouch:
I agree some usage would be nice in the README file, but this PR only mentions some ways how to use Nushell. For example, you mention only a directory module, but nupm can install scripts as well. Furthermore, the usage might change (for example, adding `nupm update` will replace `nupm install --force`).

So there could be this basic usage, but it would be good to also mention the other use cases. It doesn't have to be complete, but omitting them gives the user an inaccurate impression of nupm. (E.g., the “define a package” section could say “Nupm also supports other types of packages, see https://github.com/nushell/nupm/blob/main/docs/design/README.md#project-structure-toc for more details”).

And finally, long-term I'm hoping nupm will be self-documenting, i.e., it should be enough to just read nupm help messages in the terminal to figure out how it works. There could also be a `nupm help` command to help learning. Not the scope for this PR, just an idea.

kubouch:
OK, looks good, thanks!

PR #119, opened 2025-06-12
Replace deprecated "filter" by "where" in registry.nu

Muttsuri:
nushell deprecated filter in favour of where.
Should be drop in replacement.

## Description
After recent update a nushell message:
```console
Warning: nu::parser::deprecated

  × Command deprecated.
    ╭─[/home/muttsuri/.config/nushell/modules/nupm/utils/registry.nu:71:45]
 70 │             # Find all packages matching $package in the registry
 71 │             let pkg_files = $registry.reg | filter $name_matcher
    ·                                             ──────────┬─────────
    ·                                                       ╰── filter was deprecated in 0.105.0 and will be removed in a future release.
 72 │
    ╰────
  help: `where` command can be used instead, as it can now read the predicate closure from a variable
  ```
  I think in the relevant file nupm already uses a closure to hold the condition, so as far as I can tell no further changes should be required.
  Sorry if the PR is lackin in something I'm doing this mid work and I'm very green at this (first PR actually).


cptpiepmatz:
Oh, the CI runs on an old version of nu

fdncred:
yes

PR #121, opened 2025-06-13
docs: add examples documentation to subcommands

mkatychev:
## Description
* Added [`@example`](https://github.com/nushell/nushell/pull/14906) attribute to `nupm` and subcommands
* Fixed incorrect handling of registry `path` value in git installs
* renamed default registry `nupm_test` to `nupm` to avoid confusion with the [nupm test registry](https://github.com/nushell/nupm/blob/f4c4007531a16ecc07ba6f829e46316046b26564/tests/mod.nu#L10-L13)

### DIFF
--- a/nupm.nuon
+++ b/nupm.nuon
@@ -1,7 +1,7 @@
-{
-    name: nupm
-    type: module
-    version: "0.2.0"
-    description: "Nushell package manager"
-    license: "LICENSE"
+
+    name: nupm,
+    type: module,
+    version: "0.2.0",
+    description: "Nushell package manager",
+    license: "LICENSE",
 }
--- a/nupm/install.nu
+++ b/nupm/install.nu
@@ -153,7 +153,7 @@ def download-pkg [
     let pkg_dir = if $pkg.path == null {
         $env.PWD | path join $clone_dir
     } else {
-        $env.PWD | path join $clone_dir ($pkg.path | path dirname)
+        $env.PWD | path join $clone_dir $pkg.path
     }
 
     if ($pkg_dir | path exists) {
@@ -233,7 +233,7 @@ def fetch-package [
 # 1. Fetching the package (if the package is online)
 # 2. Installing the package (build action, if any; copy files to install location)
 export def main [
-    package  # Name, path, or link to the package
+    package # Name, path, or link to the package
     --registry: string@complete-registries  # Which registry to use (either a name
                                             # in $env.NUPM_REGISTRIES or a path)
     --pkg-version(-v): string  # Package version to install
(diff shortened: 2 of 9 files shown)


nupm/mod.nu
```diff
@@ -28,13 +28,28 @@ export-env {
...
+@example "Install a package from a local directory" { nupm install my-package --path }
+@example "Publish a package" { nupm publish my-registry.nuon --local --save }
+@example "Search for specific version" { nupm search my-package --pkg-version 1.2.0 }
+@example "Check status of specific package directory" { nupm status ./my-package }
+@example "Run tests" { nupm test }
+export def main [subcommand?]: nothing -> nothing {
     nupm-home-prompt --no-confirm=false
 
+    let subcommands = help modules | where name == nupm | get submodules.0.name
+    print $"(ansi green)Usage(ansi reset): nupm \(($subcommands | str join '|'))"
```
mkatychev, line 52: This prints out a shortform helpdoc:

```
$ nupm

Usage: nupm (install|publish|search|status|test)
enjoy nupm!
```
There's still issues with `--help` documentation for moduels (hence the `subcommand?` as well):
https://github.com/nushell/nushell/issues/10707

The funny thing is that a tab completion event on `nupm` actually displays the submodule helpdocs correctly but that's only available in the console :(

image

nupm/install.nu
```diff
@@ -153,7 +153,7 @@ def download-pkg [
     let pkg_dir = if $pkg.path == null {
         $env.PWD | path join $clone_dir
     } else {
-        $env.PWD | path join $clone_dir ($pkg.path | path dirname)
```
mkatychev, line 156: this previously failed on `nupm install nu-clippy --registry nupm_test`

mkatychev: Sorry to hear that, @cptpiepmatz @fdncred are you able to take a look at @Rydwxz's PR?

nupm.nuon
```diff
@@ -1,7 +1,7 @@
-{
-    name: nupm
-    type: module
-    version: "0.2.0"
-    description: "Nushell package manager"
-    license: "LICENSE"
+
+    name: nupm,
```
fdncred, line 2: These changes to nupm.nuon aren't necessary

fdncred (changes requested):
I'm fine with these changes except the change to nupm.nuon. Remove those changes and we can land this.

kubouch:
The `nupm_test` was named as such to denote that it's just for testing. It's not an official registry, as nupm itself is still largely incomplete and prone to change. Could be renamed to something else, though (or the testing registry can be renamed).

The `filter` I just fixed, let's not mix unrelated changes.

I think you can just trim the changes down to just adding the examples and the PR would be fine. If you want to rename the registries, you can make a new PR.

nupm.nuon
mkatychev: @kubouch is there a spec for nuon format? I thought it was a superset of json, I looked here and could not find a bnf or spec documentation other than test cases:
https://github.com/nushell/nushell/tree/main/crates/nuon

PR #122, opened 2025-06-13
feat: `nupm registry` subcommand

mkatychev:
This PR introduces the `registry` subcommand for nupm to allow registry management and support multiple registries:

```
Usage:
  > nupm registry

Subcommands:

  nupm registry add - Add a new registry
  nupm registry describe - Show detailed information about a specific registry
returning a list of package names, type, and version
  nupm registry fetch - Fetch and cache registry data locally
  nupm registry init - Initialize a new nupm registry or a registry index if the `--index` flag is
passed in
  nupm registry list - List all configured registries
  nupm registry remove - Remove a registry
  nupm registry rename - Rename a registry
  nupm registry set-url - Update a given registry url
```


mkatychev:
@kubouch thanks for the feedback, do note this is a draft PR so

> * What is the difference between registry add and registry new that you have in the TODO

I could use some feedback here but I see three usecases when dealing with registries that aren't covered yet:
* initialize a new registry (`registry new`? perhaps `registry init` is better)
* add a registry to the registry index (`registry add`)
* initialize a package (`package init`?)

kubouch:
I'd say for consistency, if we initialize something, let's call it `init`. Hence:
* Initialize a new registry: `nupm registry init`
* Add a registry to the index: `nupm registry add`
* Initialize a new package: `nupm init` (by default `nupm` operates on packages so no need to add a `package` IMO)

Rydwxz:
> Seems like you're adding a new concept of "registry index" file that replaces the `$env.NUPM_REGISTRIES`. I think that's fine, some thoughts:

I think we should prefer environment variables over files configuration for nupm, for these reason:
- there should be one consistent method to configure all the different parts of nupm
- that way should be as close to the same way nushell itself is configured as reasonably possible, not hidden in nupm internals
- the contents of $NUPM_HOME should be managed by nupm, not the user, and we should not assume that it will always be inside $XDG_CONFIG_HOME
- allows it to be set by parent processes for security reasons without complicated secondary hierarchical loading system

Since we may collect a few more variables, a nested record like this could simply be added to the end of config.nu:
```$env.nupm_config = { 
    nupm_home: "/home/rw/.config/nushell/nupm" # can still be a default
    registries: { 
        nupm: "github.com/nushell/nupm/registry.nuon" 
        awesome_nu: "..." 
    }
    sync_list: { ... }
}
```

mkatychev:
@Rydwxz are you suggesting registries to be added manually?

nupm/utils/dirs.nu
```diff
@@ -90,7 +75,7 @@ export def cache-dir [--ensure]: nothing -> path {
 }
 
 export def tmp-dir [subdir: string, --ensure]: nothing -> path {
-    let d = $env.NUPM_TEMP
+    let d = $env.nupm.temp
```
mkatychev, line 78: this is giving me a failure in `tests [ with-test-env ]`, scratching my head at this:

Error: nu::shell::column_not_found

  x Cannot find column 'nupm'
    ,-[/Users/mkatychev/.config/nupm/nupm/utils/dirs.nu:80:13]
 79 | export def tmp-dir [subdir: string, --ensure]: nothing -> path {
 80 |     let d = $env.nupm.temp
    :             ^^^^^^^|^^^^^^|
    :                    |      `-- value originates here
    :                    `-- cannot find column 'nupm'
 81 |         | path join $subdir
    `----

Ran 1 tests. 0 succeeded, 1 failed.
Error:   × some tests failed
```

Comment (mkatychev) [reply nupm/utils/dirs.nu:78]: @Rydwxz @kubouch   I've managed to reproduce the issue, it's quite strange, take a look here:
https://github.com/nushell/nushell/issues/16036

Comment (mkatychev) [reply nupm/utils/dirs.nu]: Workaround implemented in https://github.com/nushell/nupm/pull/122/commits/1f1249098165c3d1ab25446642e4c769a627225a

**Comment by mkatychev** [comment]
@Rydwxz  @kubouch  this is ready for review. Since there's already a lot to review I would like to defer commands to initialize registries and packages into a separate PR:

https://github.com/nushell/nupm/blob/baaf4f42103ade7634bd15dab981a288773e3279/nupm/registry.nu#L241-L251

File: nupm/registry.nu [hunk 2]
```diff
@@ -0,0 +1,252 @@
...
+    if $all {
+        # Fetch all registries
+        let registries = $env.nupm.registries | transpose name url
+        print $"Fetching ($registries | length) registries..."
+
+        $registries | each {|reg|
+            fetch-registry $reg.name $reg.url
+        }
+
+        print "All registries fetched successfully."
```
mkatychev, line 171: should these be `log debug`s instead of `print`?

tests/mod.nu
```diff
@@ -171,3 +169,211 @@ export def generate-local-registry [] {
...
+
+        # Test error cases
+        let no_name_result = try {
+            nupm registry fetch
+            "should not reach here"
+        } catch {|err|
+            $err.msg
+        }
+
+        assert ("Please specify a registry name or use --all flag" in $no_name_result)
```
mkatychev, line 368: I could use some suggestion on how to make this essentially do `(nupm registry fetch).is_err()`

nupm/registry.nu
```diff
@@ -0,0 +1,252 @@
...
+
+    if not ($registry in $env.NUPM_REGISTRIES) {
+        throw-error $"Registry '($registry)' not found"
+    }
+
+    let registry_url = $env.NUPM_REGISTRIES | get $registry
+    let registry_cache_dir = cache-dir --ensure | path join $registry
+    let cached_registry = $registry_cache_dir | path join "registry.nuon"
+
+    try {
```
kubouch, line 36: The mega-`try` is kind of nuclear solution. It's better to handle errors where they appear because now this doesn't tell you much more than "It failed" without saying what and how to fix it.

nupm/registry.nu
```diff
@@ -0,0 +1,252 @@
...
+      $env.NUPM_REGISTRIES | save --force $env.NUPM_INDEX_PATH
+    }
+
+    print $"Registry '($name)' renamed successfully."
+}
+
+# Fetch and cache registry data locally
+@example "Fetch a specific registry" { nupm registry fetch nupm }
+@example "Fetch all registries" { nupm registry fetch --all }
+export def fetch [
```
kubouch, line 158: So this fetches `registry.nuon` and `<package>.nuon` files of the desired registries into cache? Why do we need to cache registries? It introduces a problem if the original registry is updated, how does nupm know to update the cache? We should cache downloaded packages with proper hashing etc., but looking up a registry is just fetching a single file, it shouldn't be a problem and attempting to cache it would introduce unnecessary problems IMO.

kubouch (changes requested):

docs/design/README.md
```diff
@@ -1,8 +1,8 @@
-# Design of `nupm` :warning: Work In Progress :warning:
```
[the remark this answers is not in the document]
mkatychev: Apologies, thought this was a squash merge repo, I will reword

mkatychev:
@kubouch hopefully this should address most of the issues.

docs/design/README.md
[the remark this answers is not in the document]
kubouch: Sorry, I meant, could you move the .md files into a separate PRs? These changes are unrelated to adding the `nupm registry` subcommand. Yes, we squash, which would result in these typos being labeled as "nupm registry subcommand" -- this can be annoying eg. when using Git blame.

nupm/registry.nu
```diff
@@ -0,0 +1,178 @@
...
+    registry: string@registry-names
+]: nothing -> table {
+    use utils/dirs.nu cache-dir
+
+    if not ($registry in $env.NUPM_REGISTRIES) {
+        throw-error $"Registry '($registry)' not found"
+    }
+
+    let registry_url = $env.NUPM_REGISTRIES | get $registry
+    let registry_cache_dir = cache-dir --ensure | path join $registry
```
kubouch, line 33: I think you're missing the `registry` subdirectory here. It now dumps the files into `$env.NUPM_CACHE/<registry-name>` instead of `$env.NUPM_CACHE/registry/<registry-name>`.

nupm/registry.nu
```diff
@@ -0,0 +1,178 @@
...
+
+# Initialize a new nupm registry or a registry index if the `--index` flag is
+# passed in
+@example "Initialize registry index" { nupm registry init --index }
+export def init [--index] {
+    if $index {
+        init-index
+        return
+    }
+    # TODO initialize registry index here
```
kubouch, line 176: Here it should print something like "nupm registry init without --index is not implemented yet".

kubouch:
OK, I think it's good, except a few minor comments. We can merge it with the NUPM_REGISTRIES and refactor the system as we implement the overlays.

I originally thought that it would be better to defer implementing overlays until the base features work reliably well, but given how central overlays are to the overall design, I'm thinking that should be the next step to implement.

nupm/registry.nu
mkatychev: nice catch, I think there should be some followup tests on this

nupm/registry.nu
```diff
@@ -0,0 +1,178 @@
...
+    if ($name in $env.NUPM_REGISTRIES) {
+        throw-error $"Registry '($name)' already exists. Use 'nupm registry update' to modify it."
+    }
+    $env.NUPM_REGISTRIES = $env.NUPM_REGISTRIES | insert $name $url
+
+    if $save {
+      $env.NUPM_REGISTRIES | save --force $env.NUPM_INDEX_PATH
+    }
+
+    print $"Registry '($name)' added successfully."
```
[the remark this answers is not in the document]
mkatychev: Makes sense

nupm/registry.nu
```diff
@@ -16,6 +16,16 @@ export def list []: nothing -> table {
...
 }
 
+# Formatting convenience function that is passed in the presence of the `--save` flag, informing the user on
+# how to make their changes permanent
+export def success-msg [success_msg: string, write_path: path, saved_to_disk: bool]: nothing -> string {
+  if $saved_to_disk {
+    return $"($success_msg) and written to ($write_path)"
+  }
+
+  $success_msg | append-help "To commit the change to disk, use the `--save` flag."
```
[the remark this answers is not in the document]
kubouch: Uh, sorry, I can't read, looks good!

kubouch (approved):

PR #136, opened 2026-07-07
Migrate to New Commands

neurolag:
## Description
In nushell/nushell#18364, `str downcase` has been deprecated in favor of `str lowercase`.
Changes made in this PR will change the code accordingly and thus fix #132.

This PR will, however, break support for nushell v0.113, as the `str lowercase` command doesn't exist before nushell v0.114.0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
 
       - uses: hustcer/setup-nu@v3.10
         with:
-          version: "0.110.0"
+          version: "0.114.0"
 
       - name: Show Nushell Version
         run: version
--- a/nupm/utils/dirs.nu
+++ b/nupm/utils/dirs.nu
@@ -62,13 +62,13 @@ export def nupm-home-prompt [--no-confirm]: nothing -> bool {
 
     mut answer = ''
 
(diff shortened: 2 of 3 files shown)


fdncred:
I think you'll have to bump the ci to use 114 first

neurolag:
I'll do it tomorrow
Thank you so much for your quick responses 😄

neurolag:
Huh... now that's strange
Doesn't seem to work on Windows.

I'll work on it next week!

neurolag:
Alright, then
I did some digging and found out the error has been introduced here: nushell/nushell#17653

Because of this change, implicitly created module paths by adding `mod.nu` to a module directory will have their path separator replaced with `/`.
`nupm`s tests, however, expect paths with backslashes.

I'll change the test so it also accepts slashes.

But I think I'll also go create a PR in `nushell` because normally having files have backslashes but having implicit `mod.nu` files have forward slashes just seems inconsistent to me - I hope the `nushell` team will agree on this.

I'll push the updated test in a few mins.

fdncred:
Thanks. Consistency is usually best.

neurolag:
Awesome! Thank you so much 😄

PR #141, opened 2026-07-20
Better Linux default dirs example, formatting

nextgenthemes:
Better example for Linux

## Description
`$env.XDG_DATA_HOME` is not even set by default

Here is some AI copy pasta:

**XDG_DATA_HOME is not set by default on most standard Linux distributions** (such as Ubuntu, Fedora, or Arch); the environment variable is typically unset unless a user manually defines it or a specific sandboxed environment (like **Flatpak**) overrides it.

According to the **Freedesktop.org XDG Base Directory Specification**, if the variable is unset or empty, applications are expected to use the default path **$HOME/.local/share** for storing user-specific data files.

*   **Linux/BSD**: The variable is generally **unset by default**. It is only set if explicitly configured in shell profiles (e.g., ~/.profile, ~/.bashrc) or by desktop environments and sandbox runtimes like **Flatpak**.
*   **macOS**: The specification maps this role to **$HOME/Library/Application Support** rather than using the Linux-style `$HOME/.local/share` structure.
*   **Windows**: There is no direct XDG equivalent; Windows typically uses **%AppData%** or **%LocalAppData%**.

Therefore, the XDG Base Directory environment variables, including **XDG_DATA_HOME**, are primarily a **Linux/Unix** convention designed to declutter the home directory, and they are not active or set by default in standard desktop Linux installations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,11 @@
 > If that is the case, consider building Nushell from the `main` branch, or installing the [nightly build](https://github.com/nushell/nightly).
 
 `nupm` is a module. Download the repository and treat the [`nupm`](https://github.com/nushell/nupm/tree/main/nupm`) directory as a module. For example:
-* `use nupm/`
-* `overlay use nupm/ --prefix`
+
+```nushell
+use nupm/
+overlay use nupm/ --prefix
+```
 
 Both of the above commands will make `nupm` and all its subcommands available in your current scope. `overlay use` will allow you to `overlay hide` the `nupm` overlay when you don't need it.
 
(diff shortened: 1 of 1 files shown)


fdncred:
Why do we care about setting this to a linux default when this is supposed to be cross platform?

nextgenthemes:
Why do you put a variable in the example README that is not set, as in not working, as in broken on Linux in the readme?

Your own readme is not cross-platform! Why do you not pull the PR and fill in the info for the other OSs in the examples. Yes I put Linux in because that is what I am using and that is what I tested and that is what actually worked instead of **errors!**

I guess I am done with contributing before I even started. You really do not like help it seems, everything is always wrong.

My edit, as opposed to yours is Linux only but at least it's not broken. I would be happy to improve further, but I guess this ends here. 

I care about things working and if something is not, I care to help others to not run into broken bullshit, this seems like a foreign concept to you.

So you are a mac bro then and its working on mac? And you care only about mac? Is this it? You could have filled it back on and mentioned that this variable is working on mac if you tested that! Instead of saying anything helpful I get interrogated about proposing helpful changes, nice one.

My AI text says its "XDG_DATA_HOME, are primarily a Linux/Unix convention" so this is why I used a Linux example! Is this not self-explanatory?

fdncred:
> ... and fill in the info for the other OSs in the examples.

Good idea. Someone should do that.

nextgenthemes:
That someone could be you or who ever wrote the readme section. Again, my assumption is that this is working on Mac, I do not own apple shit so I can not easily test that.

I have the opinion it's better to have one platform example that is explicitly marked as that platform that actually works! And not literally broken, just thrown in there lazily without the mention of any platform, one variable that is "primary a Linux thing" but also broken on Linux?? How is that better? But hey you are the man, I must be doing one of my evil opinion things again.

